PR #47, opened 2026-02-07
fix(cli): remove PatchUnionIndex::build from patches list command

mmaechtel:
## Problem

`xearthlayer patches list` hangs after listing all patches when the patches directory is large (740 entries in my case).

## Cause

After printing the list, `PatchUnionIndex::build()` recursively walks every file in every valid patch folder — potentially millions of filesystem ops — just to print one summary line.

## Fix

Removed the `PatchUnionIndex::build()` block and unused import. `patches list` now returns immediately after output.

## Side effects

Not fully sure if there are downstream dependencies I'm missing. From what I can tell:
- FUSE mounting builds its own union index separately in `union_fs.rs` — unaffected
- The removed output line was purely informational
- Would appreciate review for anything I might have overlooked

## Test plan
- [ ] `patches list` with 700+ patches — should return immediately
- [ ] `xearthlayer run` — FUSE mounting still works
- [ ] `cargo check` / `cargo test` pass

### DIFF
--- a/xearthlayer-cli/src/commands/packages/args.rs
+++ b/xearthlayer-cli/src/commands/packages/args.rs
@@ -170,6 +170,8 @@ pub struct UpdateArgs {
     pub install_dir: PathBuf,
     pub temp_dir: PathBuf,
     pub all: bool,
+    /// Path to X-Plane Custom Scenery directory for overlay symlinks.
+    pub custom_scenery_path: Option<PathBuf>,
 }
 
 /// Arguments for the remove command.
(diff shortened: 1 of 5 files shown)


samsoir:
Thanks for the contribution @mmaechtel! Your diagnosis is spot on — `PatchUnionIndex::build()` recursively walks every file in every patch folder, which is completely unnecessary for a list command. With 740 patches that's millions of filesystem ops for one informational line.

We've verified this has no downstream side effects. The FUSE mount path in `union_fs.rs` builds its own index independently during `xearthlayer run`, so removing the build from `patches list` is safe.

The branch is currently behind `main` — ~could you rebase onto `main` so CI runs against the latest?~ Github took care of that for me :+1: . Once that's green we'll merge this in.

Issue #53, opened 2026-02-07
X-Plane FUSE requests trigger network calls for pre-cached tiles


samsoir:
## Summary

X-Plane occasionally sends FUSE requests for DDS tiles that result in network downloads, even when the area should be fully pre-cached (either via patches or prior prefetch/prewarm). The requests are infrequent and short-lived, but indicate XEL is not recognizing or serving certain tiles from the existing cache/patch files.

## Context

This was reported as part of #51. The primary issue there (prewarm/prefetch ignoring patch DDS tiles) is being addressed separately. This issue tracks the distinct behavior where X-Plane requests tiles that XEL doesn't expect or doesn't have cached, despite the surrounding area being fully covered.

## Observed Behavior

- Area is fully covered by zOrtho patches with DDS tiles
- Prewarm and prefetch have completed
- X-Plane still occasionally triggers FUSE requests that result in network downloads
- Downloads are short-lived but shouldn't be happening at all

## Possible Causes (to investigate)

- Tiles at the edges of zOrtho coverage that don't have DDS files (boundary effects)
- Zoom level mismatches between what X-Plane requests and what's in patches
- Tile coordinate rounding at DSF grid boundaries
- Cache key mismatch between what was cached and what X-Plane requests

## Reproduction

- Link zOrtho scenery into `patches/` with full DDS texture coverage
- Configure `scenery.ini` with zOrtho entries above the FUSE XEarthLayer directory
- Start XEL and fly around the pre-cached area
- Observe occasional network download activity in logs

Related: #51

PR #54, opened 2026-02-07
fix(prewarm): check disk for existing DDS tiles before generating

### LINKED ISSUE #51: Excessive tile processing/prefetching despite DDS tiles present in patches
## What I'm seeing

I have my own zOrtho scenery linked into `patches/`, including all pre-built DDS files. When I start XEarthLayer, two things surprise me:

1. **Prewarm runs extensively** — even though the airport area is fully covered by my linked patches with DDS tiles already in place. I would expect prewarm to detect these existing tiles and skip them.

2. **Prefetch generates a lot of tile processing in the background** — even though I have tiles linked in patches all around the airport area. I see significant download and processing activity for tiles that should already be available on disk.

3. What also surprises me is that xplane still seems to send requests from time to time, even though I have all these zOrtho tiles in my scenery.ini

## What I'd expect

If DDS tiles are already present in my `patches/` directories, neither prewarm nor prefetch should re-download or re-generate them. The existing files should be detected and skipped.

## My setup

- Custom zOrtho scenery symlinked into `patches/` (with full DDS textures)
- Tiles cover the airport and surrounding area (lots of tiles around)
- scenery.ini with all these zOrthos above FUSE xearthlayer dir

## Question

I thought the prewarm/prefetch function checked for existing DDS files in the patches before processing them. Is there something I might be doing wrong with my patch setup?

samsoir:
## Summary

Fixes #51 — the prewarm system was re-generating DDS tiles on cold start even when they already existed on disk in `patches/` or the disk cache.

- **Root cause:** `PrewarmContext::run()` only checked the in-memory cache (empty on cold start) before submitting tiles for generation. The `OrthoUnionIndex` — which knows about all on-disk DDS files including patches — was never consulted.
- **Fix:** Thread `OrthoUnionIndex` through the prewarm call chain and add a disk-existence filter (`dds_tile_exists()`) after the memory cache check, mirroring the pattern already used by the prefetch coordinator.
- **New field:** `PrewarmStatus.disk_hits` tracks tiles skipped because they exist on disk, visible in TUI progress and completion logs.

### Files changed

| File | Change |
|------|--------|
| `xearthlayer/src/prefetch/prewarm/context.rs` | Add `disk_hits` to status, thread `OrthoUnionIndex`, disk filter in `run()`, 2 new tests |
| `xearthlayer/src/prefetch/prewarm/mod.rs` | Update doc example |
| `xearthlayer/src/service/prewarm.rs` | Pass `ortho_index` through call chain |
| `xearthlayer-cli/src/tui_app.rs` | Include `disk_hits` in progress tracking |

### Note on X-Plane FUSE requests

The reporter also observed occasional FUSE requests triggering network calls despite full patch coverage. This is a separate issue tracked in #53.

### Automated
- [x] Existing prewarm unit tests updated for new `OrthoUnionIndex` parameter
- [x] `test_prewarm_skips_tiles_on_disk` — verifies tiles on disk are filtered out and counted as `disk_hits`
- [x] `test_prewarm_all_tiles_on_disk` — verifies prewarm completes immediately when all tiles exist on disk
- [x] `make pre-commit` passes (fmt + clippy + all 2022 tests)

### Manual testing

To verify the fix end-to-end:

1. **Setup**: Link zOrtho scenery with pre-built DDS tiles into `~/.xearthlayer/patches/` (symlinks work)
2. **Clear memory cache**: Restart XEarthLayer to ensure a cold start (empty memory cache)
3. **Start with prewarm**: Run `xearthlayer run` with an airport ICAO that has patch coverage (e.g., the airport your zOrtho covers)
4. **Observe TUI**: The prewarm progress bar should show rapid completion — most tiles should be counted as hits rather than needing generation
5. **Check logs**: Look for the `Prewarm filter complete` log line which now shows `cache_hits`, `disk_hits`, and `to_generate` counts. With full patch coverage, `disk_hits` should be high and `to_generate` should be low or zero
6. **Verify no redundant downloads**: Network activity during prewarm should be minimal if patches cover the prewarm area

**Expected before fix:** Prewarm shows many tiles being generated, significant network activity despite patches present.
**Expected after fix:** Prewarm shows most tiles as disk hits, minimal or no network activity for covered areas.

### DIFF
--- a/xearthlayer/src/fuse/fuse3/ortho_union_fs.rs
+++ b/xearthlayer/src/fuse/fuse3/ortho_union_fs.rs
@@ -597,14 +597,12 @@ impl Filesystem for Fuse3OrthoUnionFS {
             }
         }
 
-        // Patched region passthrough gate for DDS generation
-        // If this resource is in a region owned by a patch, the patch is responsible
-        // for providing it. Serve only what exists on disk — no generation.
-        if self.is_geo_filtered(&name_str) {
-            return Err(Errno::from(libc::ENOENT));
-        }
-
-        // Check if it's a DDS file we can generate
+        // Check if it's a DDS file we can generate.
+        //
+        // In patched regions, patch DDS files are served via resolve_lazy_geo() above
+        // (passthrough from the patch source). If we reach here, no source has this DDS
+        // on disk — generate it. This handles boundary cases where package `.ter` files
+        // reference DDS textures the patch doesn't provide (different naming convention).
         if name_str.ends_with(".dds") {
             if let Ok(coords) = parse_dds_filename(&name_str) {
                 let inode = self.inode_manager.create_virtual_inode(coords);
@@ -1338,20 +1336,32 @@ mod tests {
     /// When a patch owns a region (per GeoIndex), FUSE should never generate
     /// DDS textures for that region — it should return ENOENT for missing files.
     #[tokio::test]
-    async fn test_lookup_patched_region_returns_enoent_for_missing_dds() {
+    /// Test that DDS from a patch source is served via passthrough in patched regions.
+    ///
+    /// When a patch has the DDS file on disk, `resolve_lazy_geo()` serves it directly
+    /// (passthrough). DDS generation is never needed.
+    async fn test_lookup_patched_region_serves_patch_dds_via_passthrough() {
         use fuse3::raw::Filesystem;
 
         use crate::geo_index::{DsfRegion, GeoIndex, PatchCoverage};
 
         let temp = TempDir::new().unwrap();
 
-        // Create a patch with DSF in region (33, -119)
+        // Create a patch with DSF in region (33, -119) and a real DDS file
         let patches_dir = temp.path().join("patches");
         let patch_dir = patches_dir.join("LIPX_Mesh");
         let nav_dir = patch_dir.join("Earth nav data/+30-120");
         std::fs::create_dir_all(&nav_dir).unwrap();
         std::fs::write(nav_dir.join("+33-119.dsf"), b"fake dsf").unwrap();
-        std::fs::create_dir_all(patch_dir.join("terrain")).unwrap();
+        let patch_textures = patch_dir.join("textures");
+        std::fs::create_dir_all(&patch_textures).unwrap();
+
+        // Place a DDS file in the patch using GO2 convention at zoom 18.
+        // Patches commonly use GO218 (Google GO2 at ZL18) naming.
+        use crate::coord::to_tile_coords;
+        let tc = to_tile_coords(33.5, -118.5, 18).unwrap();
+        let dds_name = format!("{}_{}_GO218.dds", tc.row, tc.col);
+        std::fs::write(patch_textures.join(&dds_name), b"fake dds data").unwrap();
 
         // Create a package so we have a textures/ directory
         let na_pkg = create_test_package(&temp, "na");
@@ -1386,27 +1396,20 @@ mod tests {
             pid: 1000,
         };
 
-        // Look up a DDS filename whose coordinates fall in the patched region (33, -119).
-        use crate::coord::to_tile_coords;
-        let tc = to_tile_coords(33.5, -118.5, 16).unwrap();
-        let dds_name = format!("{}_{}_BI16.dds", tc.row, tc.col);
-
-        // Verify this filename actually maps to the patched region
+        // Verify this filename maps to the patched region
         use crate::prefetch::tile_based::DsfTileCoord;
         let dsf = DsfTileCoord::from_scenery_filename(&dds_name).unwrap();
         assert_eq!(dsf.lat, 33, "DDS filename should map to lat 33");
         assert_eq!(dsf.lon, -119, "DDS filename should map to lon -119");
 
-        // Lookup should return ENOENT because:
-        // 1. The DDS file doesn't exist on disk
-        // 2. GeoIndex marks this region as patch-owned → no generation
+        // Lookup should succeed — the patch has this DDS on disk, served via passthrough
         let result = fs
             .lookup(req, textures_inode, std::ffi::OsStr::new(&dds_name))
             .await;
 
         assert!(
-            result.is_err(),
-            "Lookup for DDS in patched region should return ENOENT, got: {:?}",
+            result.is_ok(),
+            "Lookup for DDS from patch should succeed via passthrough, got: {:?}",
             result
         );
     }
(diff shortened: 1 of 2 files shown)


mmaechtel:
Or do I have to use the ICAO 4-letter code for link names? I can easily change them to something like DL01, DL02, and so on. Or something else with 4 or 5 digits, as long as I can mix letters and numbers :)

samsoir:
Thanks for the updates, seems like the prewarm isn't building an accurate hashmap union of what is in the patch folder that is does not need to load. I thought I had fixed the sequencing of how that index is constructed, but more work is required.

I am creating a patch to use for testing locally which should help me quickly reproduce and then fix the issue.

> Or do I have to use the ICAO 4-letter code for link names? I

If I understand the question correctly, the prewarm system only recognizes ICAO codes that are included in X-Plane 12's airport.dat file which is used for lookup of ICAO to lat/long reference. That is then referenced against what tiles are available within the Earth nav data/ folder in any of the patches and available regional scenery. The name of the link in the patch folder.

The problem is that the prewarm system is starting before the completed union of scenery is completed, due to many things happening at once to improve startup times. I need to ensure the prewarm only starts once a complete picture of the union of available scenery is created.

mmaechtel:
yes, thats the structure in my patch folder. :)

If I can help again, just send me a ping.

samsoir:
I have found the issue - there are some incorrect assumption around how to handle patches in the way resources are reconciled in the ortho union. Fix in the works.

samsoir:
ref: https://github.com/samsoir/xearthlayer/blob/bugfix/issue-51-prewarm-ignores-patches/docs/dev/tile-patches.md

Re-reading my spec for this I now realize the behavior you reported was actually by design. The Patch system was designed to always provide it's own Ortho textures to the patched scenery, in an attempt to better match it with the ortho XEarthLayer was generating. This was necessary because Apple Maps is not an option in Ortho4XP for example.

However, as I have been working through this problem I realize that it is far too complex to do this, both from an implementation perspective, but also most importantly, it isn't how end users expect this to behave.

The change is now, if the user provides a patch for any 1x1 degree tile, XEarthLayer will defer to that patch for everything. If the user screws up the patch, that is on the user.

mmaechtel:
Great :) 

Yes, I also think that anyone who patches should know what they are doing, as the name “patch” suggests. Perhaps a warning: You need to know what you are doing when you switch to the patches/ directory ;)

But does xearthlayer differ between patches with or without dds files in patches/?

mmaechtel:
I’m asking because this would make the overall extension concept coherent.
If you want to extend your own meshes — for example zOrthos without imagery inside XEL — you can do so via packages layered through naming (e.g., your own A_Alps above the EU packages). XEL will then include these tiles in its management.
If, on the other hand, you already have zOrthos including imagery, you can declare them to XEL via patches, and XEL will completely ignore those tiles in its management.

samsoir:
> So, as far as I understand the design: Preparing my own packages and placing them above other packages still offers the possibility of having my own 'mesh' data only (ortho without DDS files). For example, building my own EU airport package could be stacked above the default EU package (by naming) and offer different data, right?

> I'm asking because this would make the overall extension concept coherent. If you want to extend your own meshes — for example zOrthos without imagery inside XEL — you can do so via packages layered through naming (e.g., your own A_Alps above the EU packages). XEL will then include these tiles in its management. If, on the other hand, you already have zOrthos including imagery, you can declare them to XEL via patches, and XEL will completely ignore those tiles in its management.

You're on the right track with the two workflows, but I want to clarify how packages work because they're not arbitrary scenery folders you can layer by naming.

**XEL packages** must conform to the XEL package specification and be compiled using the `xearthlayer publish` tooling. They can come from any repository — not just XEL's own — but they need to be built to spec. XEL manages them, generates DDS imagery on-demand, and handles caching, prewarm, prefetch, etc.

**Patches** are self-contained Ortho4XP tiles (DSF + terrain + textures) placed in `patches/`. With the changes in this PR, if a patch provides a DSF for a 1x1 degree area, XEL now defers to that patch entirely — no DDS generation, no prewarm, no prefetch for that region. This is a change from the previous behavior where XEL would attempt to generate its own DDS textures for patched areas (as described in the original tile-patches spec). Your bug report helped make it clear that the original design didn't match how users actually use patches.

So the two extensibility paths are:

1. **You want custom mesh with XEL-managed imagery** — compile your Ortho4XP output into a package using `xearthlayer publish` and include it as a managed package. XEL handles the DDS generation.
2. **You already have a complete tile (mesh + imagery)** — place it in `patches/`. XEL hands off completely for that region.

I recognize that path 1 isn't the simplest workflow for a one-off airport mesh today. Making that easier could be a future improvement, but it's out of scope for this fix.

samsoir:
I am busy testing the fix for this issue locally, hope to have the latest update pushed tomorrow (TZ: America/Los_Angeles)

samsoir:
@mmaechtel — the latest changes on this branch address the original issue you reported and the crash I mentioned in my last update. Here's a summary of what changed and how to test.

### What's fixed

The branch now includes a **GeoIndex** — a geospatial reference database that tracks which 1x1 degree regions are owned by patches. This replaces the previous filename-based approach that couldn't handle the fundamental mismatch between patch and package naming conventions (e.g., your `GO218` tiles vs XEL's `BI16` tiles covering the same area).

**Key behaviors:**

1. **Prewarm/Prefetch**: Tiles in patch-owned regions are skipped entirely — no downloads, no DDS generation
2. **FUSE**: Patch files are served via passthrough. At region boundaries where a patch doesn't cover everything, XEL falls through to generate textures on-demand from its configured provider
3. **GO2 support**: Fixed a parsing bug where `GO218` filenames were incorrectly parsed (zoom interpreted as 218 instead of 18), which affected geo-filtering for GO2-based patches

### How to test

1. Build and install the branch: `make install`
2. Place your zOrtho tiles in `~/.xearthlayer/patches/` (symlinks work). The structure should be:
   ```
   patches/
   └── A_DL_+48+09/
       ├── Earth nav data/
       │   └── +48+009/
       │       └── +48+009.dsf    ← this is what XEL uses to determine ownership
       ├── terrain/
       └── textures/
   ```
3. Start XEL: `xearthlayer run --airport LIPX` (or whichever airport your patches cover)

### What to expect

- **During startup**: You should see log messages about patch regions being registered in the GeoIndex. The count should match the number of 1x1 degree DSF files in your patches.
- **Prewarm**: Should complete quickly. Tiles in patched regions show as `patch_skipped` rather than being queued for generation. You should see minimal or zero network activity for areas covered by your patches.
- **In the sim**: Your patch scenery loads normally via passthrough. At the edges of your patched regions, you **will** see XEL generating DDS textures on-demand — this is expected and correct. These are boundary tiles where your patch's DSF doesn't extend but the regional package's terrain references textures in that area. XEL fills these gaps by generating the textures from its configured imagery provider (Bing, Google, etc.), which is why you'll see some network activity at the boundary of your patched regions.
- **No crash**: The previous crash caused by conflicting patch/package resources in the same region is resolved.

Let me know how it goes!

mmaechtel:
I did my test flight from Italy to northern Germany again. This time, xearthlayer was running at takeoff. Prewarming and prefetching were configured, but nothing was loaded from the network and nothing was processed... it was completely “quiet” because all my orthos are linked in patches. Then, as I approached the boundaries of my orthos, it filled up the empty cache. Everything was stable until loading. I still need to optimize the parameters here and there so that the FPS doesn't drop too much during cruise when xel becomes active, but I'll get that sorted out. By the way, I've already added xearthlayer to my “XplaneOnLinux” blog :) 

->[XOL: Ortho Streaming]( https://xol.emvisio.com/en/scenery/ortho_streaming/index.html )

samsoir:
I didn't realize you were the owner of the X-Plane on Linux blog. I used that extensively when I jumped over from Windows - thank you!

PR #66, opened 2026-03-04
Boundary-driven prefetch, FUSE bandwidth, GPU/ISPC encoding & transition ramp

### LINKED ISSUE #58: Prefetch refinements: takeoff grace period and cruise band depth
## Context

During live testing of the pipeline admission control fix (#56 follow-up), two prefetch refinements were identified. Neither are blockers — the pipeline is working correctly — but they would improve the experience at flight phase transitions and DSF boundary crossings.

## 1. Takeoff Grace Period

**Problem**: On takeoff, the prefetch system fires up aggressively at the same moment X-Plane ramps up rendering (LOD shifts, flight model at full fidelity). The CPU contention causes a 1-2s stutter while the OS scheduler handles the demand spike.

**Potential approaches**:
- Delay prefetch start by 10-15s after Ground → Cruise phase transition
- Gradual ramp-up: start at reduced fraction, ramp to full over 30-60s
- CPU-aware throttling via `executor_load()` on the CPU pool specifically
- System-wide CPU load monitoring (not just our pools)

## 2. Cruise Band Misalignment — Prefetch Fills Wrong Area

**Problem**: The cruise prefetch band is significantly misaligned with the tiles X-Plane actually needs. Prefetch runs continuously (no circuit breaker interference after #59 fix), but fills tiles far behind or to the side of the aircraft, leaving the area ahead uncached. X-Plane then loads the next column on-demand, causing framerate drops.

### Evidence from live flight test (2026-02-25, LOWI → LFPG, #59 branch)

**Observation 1 — 10°E boundary (05:42)**:
- Aircraft position: 47.74°N, 10.33°E, heading 292° (west)
- Prefetch was filling ZL12 cols 2150-2157 (~10.1-10.3°E) — **behind** the aircraft
- X-Plane needed ZL12 cols 2133-2134 (~9.8°E) — **ahead** of the aircraft
- On-demand burst: 200 tiles at 05:42:35, ~835 total tiles over 14 seconds
- Result: framerate drop (no freeze, thanks to fast pipeline ~180ms/tile)

**Observation 2 — 9°E boundary area (05:48)**:
- Aircraft position: ~47.6°N, ~8.9°E, heading 259°
- Prefetch frontier still at ZL12 cols 2139-2141 (~9.9°E) and ZL14 cols 8573-8585
- Prefetch was filling ZL14 cols 8756-8783 (~11.9°E) earlier — **2° east** of aircraft
- On-demand burst: 196 tiles at 05:48:12, ~523 total over 10 seconds
- Result: another framerate drop

**Observation 3 — Stale `skipped_cached` counter**:
- `skipped_cached=5348` was frozen across many cycles
- `tiles=50` or `tiles=88` planned per cycle but actual frontier barely advanced
- The prefetch band covers 15-24 DSF tiles but many are already cached from the initial scene load, so the "new" tiles being generated are in the wrong area

### Root Cause Analysis

The prefetch band calculation has two issues:

1. **Tile selection lag**: The band includes tiles that are already cached (from scene load or previous prefetch), and the remaining uncached tiles are not well-ordered by distance from the aircraft. The system generates tiles at the edges of the band rather than prioritizing the frontier directly ahead.

2. **Band geometry vs X-Plane loading pattern**: The prefetch band covers a rectangular region of DSF tiles based on `lead_distance` and `band_width`, but within that band, tiles are not sorted by relevance to the aircraft's upcoming path. X-Plane loads in columns/rows at DSF boundaries, so prefetch should prioritize the **next column/row** the aircraft will cross.

### Current configuration (`config.rs`):
- `lead_distance: 2` — 2 DSF tiles (2°) ahead
- `band_width: 2` — 2 DSF tiles perpendicular to track
- Band starts from the **next DSF boundary**, not aircraft position

### Potential fixes:
- **Priority ordering within band**: Sort uncached tiles by distance to aircraft, nearest first
- **Increase lead_distance to 3-4°**: More buffer means more time to fill ahead tiles
- **Column-first ordering**: At DSF boundaries, prioritize tiles in the next column X-Plane will load
- **Speed-based lead scaling**: Faster aircraft → deeper lead band
- **Reduce band_width to 1**: Narrower band = fewer tiles, more focused on track

## Test Context

- Route: LOWI → LFPG (MOGT1X MOGTI P66 KPT L856 RAVED Y112 HOC MOROK PENDU JAVVU GIVRI TINIL TINI9X)
- Circuit breaker: **zero false trips** during entire flight (#59 fix verified)
- Pipeline: per-tile latency 149-189ms, prefetch running continuously
- On-demand tiles caused framerate drops but no hard freezes (pipeline fast enough)
- Two boundary crossings with on-demand bursts despite active prefetch

### LINKED ISSUE #60: [Feature]: Update notification and auto-update
### Problem Statement

When running XEarthLayer locally it is not possible to tell if the local version is up to date

### Proposed Solution

If there is a newer version available on the release page than the version running locally, inform the user when XEarthLayer starts, providing the option to;

1. (S)kip version
2. (U)pgrade to x.y.z

If the user chooses to skip, XEarthLayer continues  to run with the current version.

If the user chooses to upgrade, XEarthLayer downloads the new version and then installs it - note this will have to handle overwriting the binary of the process that kicked off the upgrade, likely will need another binary to hand upgrade.

### Alternatives Considered

N/A

### Checklist

- [x] I have checked the ROADMAP.md to see if this feature is already planned
- [x] I have searched existing issues to ensure this hasn't already been requested

### LINKED ISSUE #62: Prefetch causes stutters during takeoff — needs gradual ramp-up
## Problem

During takeoff, the prefetch system submits tiles at full capacity immediately. This creates a large burst of network and CPU load at exactly the moment X-Plane needs system resources most — scene loading is still settling, the sim is rendering new terrain as the aircraft accelerates, and GPU/CPU headroom is minimal.

The result is visible framerate stutters during the takeoff roll and initial climb.

## Observed Behavior

- Prefetch begins submitting at full rate as soon as the aircraft starts moving
- Network and CPU pools spike, competing with X-Plane's own rendering and scenery loading
- Framerate drops noticeably during takeoff roll through initial climb-out
- Once established in cruise, the system stabilizes and prefetch works well

## Expected Behavior

During the takeoff phase, prefetch should:

1. **Delay** — wait before starting prefetch after the transition from ground to takeoff
2. **Ramp up** — gradually increase submission rate rather than immediately going to full capacity
3. **Yield to on-demand** — be more conservative about resource usage while X-Plane is actively loading new scenery columns

## Context

- The cruise strategy and ground strategy both work well once the aircraft is established in their respective phases
- The issue is specifically the **transition period** during takeoff where system resources are most constrained
- Related to the adaptive prefetch coordinator's phase detection (`PhaseDetector`)
- The circuit breaker (#59) prevents permanent blocking but doesn't address the burst/stutter problem

## Enhancement

This is an enhancement to the prefetch system's phase handling. Implementation details to be discussed when we work on this issue.

### LINKED ISSUE #67: Raise FUSE max_background from 12 to support X-Plane 12.4 multithreaded loading
## Context

X-Plane 12.4 introduced a multithreaded scenery loading system. Profiling XEL's FUSE layer reveals that the kernel FUSE driver limits concurrent async requests to **12** (`max_background`), with congestion backpressure starting at **9** (`congestion_threshold`).

This is hardcoded in the `fuse3` crate (v0.8.1) at `src/raw/abi.rs:39`:

```rust
pub const DEFAULT_MAX_BACKGROUND: u16 = 12;
pub const DEFAULT_CONGESTION_THRESHOLD: u16 = DEFAULT_MAX_BACKGROUND * 3 / 4; // = 9
```

These values are sent to the kernel during `FUSE_INIT` and are **not exposed** via `MountOptions` or any public API.

## Evidence

From Chrome Trace profiling (PR #66):
- Max concurrent `fuse_read` observed: **13** (12 background + 1 foreground)
- `congestion_threshold` of 9 causes X-Plane's VFS reads to block at the kernel level
- XEL's pipeline has ample capacity — executor, HTTP, and CPU pools are all underutilized at 12 concurrent reads
- DSF boundary stalls correlate with FUSE congestion, not executor saturation

## Impact

With the pipeline improvements in PR #66 (1024 HTTP permits, 64 network pool, FD limit raising), XEL can sustain ~50 tiles/sec. But the FUSE bottleneck means X-Plane can only request 12 tiles concurrently — starving the pipeline during burst loading at DSF boundaries.

Raising `max_background` to 64–128 would let the kernel feed requests at the rate XEL can process them.

## Proposed Approach

Patch the `fuse3` crate to expose `max_background` and `congestion_threshold` as configurable session parameters:

1. **`[patch.crates-io]`** — local patch dependency in `Cargo.toml` pointing to a patched fuse3
2. Add a `max_background` setter to `MountOptions` or the `Session` builder
3. Set XEL's default to 64 (or auto-tune based on CPU count)
4. Optionally upstream the patch to the fuse3 crate

The code change in fuse3 is small — adding a field to the `fuse_init_out` construction in `session.rs` and exposing it through the public API.

## References

- Profiling trace: PR #66 (`feature/prefetch-ordering-and-ramp-58-62`)
- FUSE kernel docs: `max_background` added in FUSE protocol 7.8
- X-Plane 12.4 release notes: multithreaded scenery loading

samsoir:
## Summary

Fixes #58. Fixes #60. Fixes #62. Fixes #67.

Replaces the dead-reckoning band prefetch model with a boundary-driven approach that matches how X-Plane actually loads scenery, fixes FUSE kernel bottlenecks, adds GPU and ISPC SIMD texture compression, corrects the prefetch window dimensions, and introduces a transition throttle to eliminate takeoff stutters.

### Boundary-Driven Prefetch (#58)

X-Plane loads scenery using a position-only, boundary-triggered model with two independent monitors (latitude and longitude). The previous `CruiseStrategy` used dead-reckoning band projection that targeted the wrong regions — prefetching behind the aircraft while X-Plane needed tiles ahead.

The new model:
- **SceneryWindow** derives window dimensions from the initial scene load, tracks a three-window model (XP Window, XEL Window, Retained set) with dual boundary monitors
- **BoundaryMonitor** detects when the aircraft approaches DSF region edges on row and column axes independently
- **BoundaryStrategy** converts predicted boundary crossings into exact rows/columns X-Plane will request next
- **GeoIndex extensions** track prefetch state per region (Absent → InProgress → Prefetched → NoCoverage) with two-phase commit

Removes `CruiseStrategy`, `BandCalculator`, `BoundaryPrioritizer`, and `TurnDetector` (~1,580 lines deleted, ~1,650 lines added).

### Prefetch Window & Prewarm Corrections (#60)

Empirical measurement at three airports (EDDF 50°N, YPAD 35°S, WSSS 1°N) revealed the actual X-Plane scenery window is ~3° lat × ceil(3°/cos(lat)) lon — NOT the assumed 9×9°. The previous window was ~9× too large, causing prefetch to target distant regions.

- **Window**: Fixed to 3° latitude (constant) × dynamic longitude via `lon_tiles_for_latitude()` cosine formula
- **Load depth**: Now asymmetric — 3 rows for latitude crossings, 2 columns for longitude crossings (from 65 classified FUSE burst events)
- **Trigger distance**: Reduced from 3.0° to 1.5° (must exceed X-Plane's 1.0° trigger)
- **Prewarm grid**: Changed from 4×4 square to 3×4 (rows × cols) to match actual window
- **Config chain**: New keys `load_depth_lat`, `load_depth_lon`, `window_lon_extent`, `prewarm.grid_rows`, `prewarm.grid_cols`; old keys deprecated via `upgrade.rs`

### FUSE Bandwidth (#60)

The `fuse3` crate hardcodes `max_background=12` and `congestion_threshold=9`, severely limiting concurrent background FUSE requests. The kernel throttles at 9 in-flight requests, causing sim freezes at DSF boundaries.

- **fuse3 fork**: Extended `ReplyInit` with optional `max_background`/`congestion_threshold` fields
- **Defaults**: 256 max background / 192 congestion threshold (was 12/9)
- **Configurable**: New `[fuse]` config section with `max_background` and `congestion_threshold` settings
- **Full config chain**: `defaults.rs` → `settings.rs` → `keys.rs` → `parser.rs` → `writer.rs` → `orchestrator_config.rs` → FUSE `init()` handshake

### DDS Compression Backends (#67)

Flight testing showed CPU saturation (100% on 32 cores) at DSF boundaries during texture encoding. The `BlockCompressor` trait provides pluggable compression backends:

- **`BlockCompressor` trait**: `Send + Sync` abstraction with `compress(&self, image, format) -> Result<Vec<u8>>` 
- **`SoftwareCompressor`**: Pure-Rust block-by-block fallback
- **`IspcCompressor`** (default): Intel ISPC SIMD kernels via `intel_tex_2` — processes entire surfaces via SSE4/AVX2, 5-10× faster than software
- **`WgpuCompressor`** (feature-gated): Offloads BC1/BC3 compression to GPU compute shaders via wgpu + `block_compression` crate (ISPC kernels ported to WGSL). Targets idle integrated GPU (e.g., AMD Radeon on Ryzen APU) leaving the discrete GPU free for X-Plane rendering
- **GPU device selection**: Config-driven via `texture.gpu_device` — selects by device type (`integrated`/`discrete`) or adapter name substring. Hard error at startup if device not found (no silent fallback)
- **Feature gate**: `--features gpu-encode` required for GPU path; adds `wgpu`, `block_compression`, `pollster` deps
- **Eager initialization**: GPU adapter enumeration and device request happen at startup for fail-fast behavior
- **Diagnostics**: `xearthlayer diagnostics` lists available wgpu compute adapters when built with GPU support
- **Makefile**: `make release-gpu` and `make install-gpu` targets

Configuration:
```ini
[texture]
compressor = gpu          ; software, ispc (default), or gpu
gpu_device = integrated   ; integrated, discrete, or adapter name substring
```

### Transition Throttle (#62)

Eliminates framerate stutters during takeoff by ramping prefetch capacity gradually after the Ground → Cruise phase transition:
- **TransitionThrottle** — grace period (default 10s) + linear ramp-up (default 30s) state machine
- **PhaseDetector** — new Transition phase with MSL-based release (reaches cruise altitude threshold)
- Yields system resources to X-Plane during the critical takeoff/initial climb period

### Chrome Trace Profiling

Enables data-driven performance analysis with `--profile` flag:
- Feature-gated `tracing-chrome` integration (`cargo build --features profiling`)
- Produces Chrome Trace Format JSON viewable in [Perfetto UI](https://ui.perfetto.dev)
- Dedicated `target: "profiling"` on spans for surgical trace filtering without ambient noise
- Non-blocking trace writer to avoid mutex contention across async tasks

### HTTP Download Throughput

- **4x HTTP concurrency**: semaphore raised from 256 to 1024 concurrent requests
- **FD limit management**: `raise_fd_limit()` via `rlimit` crate raises soft limit to hard maximum at startup
- Result: ~47MB/s sustained download throughput (up from ~30MB/s)

---

## Other Changes

- **Combinable `--debug` and `--profile` flags** — both can now be used simultaneously
- **Config pipeline** — new settings for boundary-driven prefetch, FUSE limits, prewarm grid, compressor selection; deprecated band settings and old keys
- **`IntegerRangeSpec` / `FloatRangeSpec`** — new config validators for bounded numeric values
- **Flight test data** — compressed logs from 5 flights used in scenery loading whitepaper research
- **Whitepaper v1.2** — updated X-Plane scenery loading analysis with empirical measurements
- **Design doc v2.1** — adaptive prefetch design rewritten for boundary-driven architecture
- **Git pre-commit hook** — `make setup-hooks` installs a pre-commit hook that runs `make pre-commit`
- **Reverted concurrent daemon handlers** — spawned request handlers caused 50% throughput regression

## Test Plan

- [x] `make pre-commit` passes (fmt + clippy + 2,260 tests)
- [x] `cargo check --features gpu-encode` compiles cleanly
- [x] `cargo check --features profiling` compiles cleanly
- [x] All 7 GPU-specific tests pass (`cargo test --features gpu-encode wgpu`)
- [x] GPU compressor without feature flag returns hard `ServiceError::ConfigError`
- [x] Flight test at EDDH: no freezes at DSF boundaries (was 6-8s freezes)
- [x] GPU encoding test: CPU load significantly reduced, near-zero sim performance impact
- [x] iGPU (AMD Radeon integrated) faster than pipeline can feed it — compression is no longer bottleneck
- [x] Initial scene load improved from ~3m to <2m (cached start)
- [x] Takeoff test: no stutters during Ground → Cruise transition
- [x] `--profile` run produces valid trace JSON loadable in Perfetto UI
- [x] High-throughput download: no EMFILE errors with 1024 concurrent HTTP requests
- [ ] Extended flight test: verify memory cache hit rate improvement with corrected window

### DIFF
--- a/xearthlayer/src/metrics/snapshot.rs
+++ b/xearthlayer/src/metrics/snapshot.rs
@@ -51,12 +51,14 @@ pub struct TelemetrySnapshot {
     pub downloads_active: usize,
 
     // === Cache metrics ===
-    /// Memory cache hits
+    /// Memory cache hits (all origins)
     pub memory_cache_hits: u64,
-    /// Memory cache misses
+    /// Memory cache misses (all origins)
     pub memory_cache_misses: u64,
-    /// Memory cache hit rate (0.0 - 1.0)
+    /// Memory cache hit rate — all origins (0.0 - 1.0)
     pub memory_cache_hit_rate: f64,
+    /// Memory cache hit rate — FUSE (X-Plane) requests only (0.0 - 1.0)
+    pub fuse_memory_cache_hit_rate: f64,
     /// Memory cache current size in bytes
     pub memory_cache_size_bytes: u64,
     /// Disk cache hits
@@ -123,6 +125,7 @@ impl Default for TelemetrySnapshot {
             memory_cache_hits: 0,
             memory_cache_misses: 0,
             memory_cache_hit_rate: 0.0,
+            fuse_memory_cache_hit_rate: 0.0,
             memory_cache_size_bytes: 0,
             disk_cache_hits: 0,
             disk_cache_misses: 0,
@@ -223,15 +226,16 @@ impl fmt::Display for TelemetrySnapshot {
         writeln!(
             f,
             "  Completed: {} ({:.1}/s)",
-            self.jobs_completed, self.jobs_per_second
+            format_number(self.jobs_completed),
+            self.jobs_per_second
         )?;
-        writeln!(f, "  Active: {}", self.jobs_active)?;
-        writeln!(f, "  Timed out: {}", self.jobs_timed_out)?;
-        writeln!(f, "  Errors: {}", self.jobs_failed)?;
+        writeln!(f, "  Active: {}", format_number(self.jobs_active))?;
+        writeln!(f, "  Timed out: {}", format_number(self.jobs_timed_out))?;
+        writeln!(f, "  Errors: {}", format_number(self.jobs_failed))?;
         writeln!(
             f,
             "  Coalesced: {} ({:.1}%)",
-            self.jobs_coalesced,
+            format_number(self.jobs_coalesced),
             self.coalescing_rate() * 100.0
         )?;
         writeln!(f)?;
@@ -241,33 +245,35 @@ impl fmt::Display for TelemetrySnapshot {
         writeln!(
             f,
             "  Chunks: {} ({:.0}/s)",
-            self.chunks_downloaded, self.chunks_per_second
+            format_number(self.chunks_downloaded),
+            self.chunks_per_second
         )?;
         writeln!(f, "  Throughput: {}", self.throughput_human())?;
         writeln!(
             f,
             "  Failed: {} ({:.2}%)",
-            self.chunks_failed,
+            format_number(self.chunks_failed),
             self.chunk_failure_rate() * 100.0
         )?;
-        writeln!(f, "  Retries: {}", self.chunks_retried)?;
+        writeln!(f, "  Retries: {}", format_number(self.chunks_retried))?;
         writeln!(f)?;
 
         // Cache
         writeln!(f, "Cache:")?;
         writeln!(
             f,
-            "  Memory: {:.1}% hit rate ({} hits, {} misses)",
+            "  Memory: {:.1}% hit rate ({} hits, {} misses) | FUSE: {:.1}%",
             self.memory_cache_hit_rate * 100.0,
-            self.memory_cache_hits,
-            self.memory_cache_misses
+            format_number(self.memory_cache_hits),
+            format_number(self.memory_cache_misses),
+            self.fuse_memory_cache_hit_rate * 100.0,
         )?;
         writeln!(
             f,
             "  Disk: {:.1}% hit rate ({} hits, {} misses)",
             self.disk_cache_hit_rate * 100.0,
-            self.disk_cache_hits,
-            self.disk_cache_misses
+            format_number(self.disk_cache_hits),
+            format_number(self.disk_cache_misses)
         )?;
 
         Ok(())
@@ -298,6 +304,19 @@ fn format_bytes(bytes: u64) -> String {
     } else {
         format!("{} B", bytes)
     }
+}
+
+/// Format an integer with thousand separators (e.g., 23040 → "23,040").
+fn format_number(n: impl fmt::Display) -> String {
+    let s = n.to_string();
+    let mut result = String::with_capacity(s.len() + s.len() / 3);
+    for (i, c) in s.chars().enumerate() {
+        if i > 0 && (s.len() - i).is_multiple_of(3) {
+            result.push(',');
+        }
+        result.push(c);
+    }
+    result
 }
 
 /// Format duration in human-readable form.
@@ -338,6 +357,7 @@ mod tests {
             memory_cache_hits: 30,
             memory_cache_misses: 60,
             memory_cache_hit_rate: 0.333,
+            fuse_memory_cache_hit_rate: 0.6,
             memory_cache_size_bytes: 500_000_000,
             disk_cache_hits: 5000,
             disk_cache_misses: 18040,
@@ -423,6 +443,16 @@ mod tests {
     }
 
     #[test]
+    fn test_format_number() {
+        assert_eq!(format_number(0), "0");
+        assert_eq!(format_number(999), "999");
+        assert_eq!(format_number(1_000), "1,000");
+        assert_eq!(format_number(23_040), "23,040");
+        assert_eq!(format_number(1_000_000), "1,000,000");
+        assert_eq!(format_number(5_000_000_000_u64), "5,000,000,000");
+    }
+
+    #[test]
     fn test_display() {
         let snapshot = test_snapshot();
         let output = format!("{}", snapshot);
@@ -431,6 +461,10 @@ mod tests {
         assert!(output.contains("Jobs:"));
         assert!(output.contains("Downloads:"));
         assert!(output.contains("Cache:"));
+        // Verify thousand separators
+        assert!(output.contains("23,040")); // chunks_downloaded
+        assert!(output.contains("5,000 hits")); // disk_cache_hits
+        assert!(output.contains("18,040 misses")); // disk_cache_misses
     }
 
     #[test]


samsoir:
:shipit:

PR #69, opened 2026-03-12
fix(fuse): prevent virtual DDS generation for water mask textures

### LINKED ISSUE #68: Water mask PNG files may not be served correctly via FUSE
## Summary

Water/land transition artifacts are visible in X-Plane when using XEL-served scenery packages. The artifacts appear as hard rectangular (chunk-sized) boundaries at coastlines, suggesting that water mask PNG files referenced by `_sea_overlay.ter` terrain files are not being served correctly.

## Observed Behavior

- Visible rectangular chunk boundaries at water/land transitions
- Reported at coastal areas (Greece islands, France, Netherlands coastlines)
- Artifacts are consistent across different scenery packages and Ortho4XP settings

## Expected Behavior

Smooth water/land transitions using the `BORDER_TEX` PNG alpha masks that define coastline boundaries.

## Background

Ortho4XP generates three types of `.ter` files:
- `X.ter` — land tiles (`NO_ALPHA`)
- `X_sea.ter` — full-water tiles (`WATER_COLOR_MASK` + `WET`)
- `X_sea_overlay.ter` — coastal transition tiles with `BORDER_TEX ../textures/X.png`

The `_sea_overlay.ter` files reference PNG water masks via `BORDER_TEX`. These PNGs exist on disk in the scenery packages (e.g., 221,925 PNGs in the EU ortho package alone). The FUSE layer should serve them as passthrough files via lazy resolution (`textures/` is a lazy directory).

## Investigation Needed

- Confirm whether PNG `BORDER_TEX` lookups are reaching the FUSE layer
- Confirm whether `resolve_lazy_geo()` successfully resolves PNG paths
- Check if `_sea_overlay.ter` files themselves are being served
- Rule out timing/attribute issues with the DDS + PNG serving combination

## Related

- samsoir/xearthlayer-regional-scenery#1 — original user report with screenshot

samsoir:
## Summary

- Fixes water/land coastline artifacts where satellite imagery extended into water areas
- In FUSE `lookup()`, checks for a `.png` water mask on disk before generating a virtual DDS inode
- If a PNG exists, returns ENOENT so X-Plane falls back to the real water mask file

Fixes #68

## Root Cause

X-Plane resolves `BORDER_TEX` water masks by requesting `.dds` first, then `.png`. XEL's FUSE layer intercepted these `.dds` lookups and generated satellite imagery instead of letting X-Plane fall back to the actual PNG alpha mask on disk.

## Test plan

- [x] `make pre-commit` passes (fmt + clippy + 2264 tests)
- [x] Manual test: loaded at LGKR with EU ortho package — coastline water masks render correctly
- [x] Verified via FUSE diagnostic logging: X-Plane requests only `.dds`/`.ktx2` for textures, never `.png` directly

### DIFF
--- a/xearthlayer/src/fuse/fuse3/ortho_union_fs.rs
+++ b/xearthlayer/src/fuse/fuse3/ortho_union_fs.rs
@@ -1652,6 +1652,138 @@ mod tests {
         );
     }
 
+    // ========================================================================
+    // Water mask guard tests (Issue #68)
+    // ========================================================================
+
+    /// Test that lookup returns ENOENT for DDS when a PNG water mask exists.
+    ///
+    /// X-Plane requests BORDER_TEX water masks as `.dds` first. If a `.png`
+    /// with the same stem exists on disk, we must return ENOENT so X-Plane
+    /// falls back to the real PNG water mask instead of getting satellite imagery.
+    #[tokio::test]
+    async fn test_lookup_returns_enoent_for_dds_when_png_water_mask_exists() {
+        use fuse3::raw::Filesystem;
+
+        let temp = TempDir::new().unwrap();
+
+        // Create a package with a PNG water mask in textures/
+        let pkg_dir = temp.path().join("eu_ortho");
+        std::fs::create_dir_all(pkg_dir.join("Earth nav data/+40-080")).unwrap();
+        std::fs::write(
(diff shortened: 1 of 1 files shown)


samsoir:
:shipit:

Issue #176, opened 2026-04-12
prefetch: promotion correctness is downstream of scenery index completeness


samsoir:
## Problem

Region promotion (`promote_completed_regions`, `evaluate_stale_regions`) decides whether all tiles in a DSF region are cached by calling `BoundaryStrategy::tiles_for_region`, which uses `SceneryIndex::tiles_near` to enumerate the tiles that *should* exist for the region. The promotion check is then: \"are all of those tiles present in the cache?\"

If the `SceneryIndex` is incomplete — a `.ter` file failed to parse, a package wasn't fully scanned, or an entry was missed during the cache build — `tiles_for_region` returns fewer tiles than X-Plane will actually request. Promotion can then fire *prematurely* on a region where XEL believes all tiles are covered but reality has tiles the index never knew about.

Observable symptom: a region appears `Prefetched`, but X-Plane requests a tile within it that isn't in the cache, triggering a FUSE-on-demand fetch. This is indistinguishable from the prefetch-coverage bug until you trace it back to the index.

## Why this is worth tracking separately

- **Not introduced by #172 or #175.** This risk is pre-existing and orthogonal to the mark-before-submit / shadow-state defects.
- **CacheLocator (#175) doesn't fully fix it.** The locator answers \"is this *specific* tile cached?\" correctly regardless of scenery index state. But the question \"is this region complete?\" — which drives promotion decisions — still depends on the scenery index knowing the full tile set.
- **Hard to detect from logs today.** No current instrumentation distinguishes \"tile wasn't prefetched because it was outside the scenery index\" from \"tile wasn't prefetched due to other reasons.\"

## Possible approaches (for later design discussion)

1. **Scenery index self-audit.** After build, cross-reference against the FUSE-observable tile request stream. Log entries when X-Plane requests tiles not in the index (suggests a missing `.ter` parse).
2. **`.ter` parse error visibility.** Currently parse failures are logged at `trace!` level (see `scenery_index.rs:173`). Escalate to `warn!` and surface the count in startup summary.
3. **Completeness metric at startup.** Expected tile count per DSF vs. indexed tile count, flagged when anomalously low (e.g., < 75% of expected based on installed package manifests).
4. **Defensive promotion policy.** Instead of \"all tiles cached → promote\", use \"all indexed tiles cached AND no FUSE misses observed in this region → promote.\" Requires crossing prefetch with FUSE telemetry.

## Not in scope right now

[...]

Issue #175, opened 2026-04-12
refactor: unified CacheLocator SSOT — remove cache-internal leakage across prefetch/FUSE/TUI


samsoir:
## Goal

Refactor cache state resolution so the cache system is the **single source of truth** for \"is this tile available, and where?\". Today, multiple subsystems (prefetch coordinator, FUSE, TUI, executor) each maintain their own view of cache state — shadow HashSets, region state machines, retention windows — and these views drift apart because none of them reconcile against the authoritative cache contents. The result is a class of bugs where one subsystem believes a tile is cached while another has already evicted it.

#172 is the most recent symptom. The narrow fix merged there addresses the specific mark-before-submit failure mode, but it leaves the shadow state (`cached_tiles` HashSet, `PrefetchedRegion::Prefetched` state, retention machinery) intact. Those structures exist *because* there's no unified cache query interface — they're local caches-of-caches built to avoid repeatedly querying the real cache.

This issue tracks the architectural cleanup that eliminates them.

### Introduce a `CacheLocator` trait in `xearthlayer/src/cache/`

Single interface, consumed by prefetch, FUSE, TUI, and the executor:

```rust
pub trait CacheLocator: Send + Sync {
    fn locate(&self, tile: &TileCoord) -> CachePresence;
    // CachePresence::{Memory, DdsDisk, Package, Absent}
}
```

One query, one answer, derived from the actual cache tier state. No shadow, no drift.

### Remove shadow state that currently exists only to bound queries against the real cache

- **`cached_tiles: HashSet<TileCoord>`** in `AdaptivePrefetchCoordinator` — every caller becomes a `CacheLocator::locate` call. Fast-path optimisation inside `filter_memory_cache` is replaced with the direct locator query (O(1) already).
- **`PrefetchedRegion::Prefetched` state** — region completion derives from \"are all its tiles resolvable via `CacheLocator` to a non-Absent value?\". The `Prefetched` variant of the enum becomes redundant; keep `InProgress` and `NoCoverage` only.
- **Retention machinery** (`RetainedRegion` layer, `update_retention*`, `evict_non_retained`, `evict_cached_tiles_outside_retained`) — retention exists to bound shadow-state growth. With no shadow, there is nothing to retain. LRU on the DDS disk cache already discards the right tiles; retention's \"keep these specific regions\" signal has no value beyond LRU.

### Simplify downstream callers

[...]

PR #177, opened 2026-04-13
fix(prefetch): complete #172 — DDS cache keys, unified box, debug map fidelity

### LINKED ISSUE #172: prefetch: coverage degrades on long-haul flights, tiles not ready on approach
## Summary

Prefetched tiles around the departure airport fail to make it into the cache before X-Plane requests them, causing FUSE to fetch tiles on-demand over the network instead of serving them from the DDS disk cache. The bug is diagnostically specific: **regions committed to the coordinator's `InProgress` state before their tiles actually reach the executor**, combined with the Transition-throttle ramp dropping ~80% of the planned plan in the first 31 seconds of cruise.

Symptom reported: visible tile coverage gaps around the airport on a long-haul AMS→MSP test flight; confirmed in a second LOWW westbound flight (9h log, 2.8GB, debug instrumentation branch).

## Root cause (verified against flight log — see comment thread)

1. **`mark_in_progress` fires *before* `execute_plan`.** Every region that survives `max_tiles_per_cycle` truncation is committed as `InProgress` regardless of how many of its tiles actually get submitted.
2. **Transition throttle ramps 10% → 100% over 31 seconds** at Ground → Cruise. In that window, ~555 tiles were discarded across ~5,000 planned. No `Channel full`, no backpressure defer — pure throttle-ramp loss.
3. **`pending_tiles` overflow is replaced, not merged, across cycles** (core.rs:664: `self.pending_tiles = result.pending`). Partial-drain cycles leak leftover tiles.
4. **`promote_completed_regions` consults a shadow `HashSet` (`cached_tiles`) rather than the actual DDS disk cache.** Flight log shows 4 normal promotions vs. 61 "stale rescue" promotions (via `evaluate_stale_regions` which does check the real cache). The shadow state is failing to track ~94% of actually-cached tiles; the rescue path is carrying the system.

Net effect: during the first ~30 seconds of cruise, 12 contiguous DSFs forming the departure bubble each have ~200–300 tiles planned, ~20–40 submitted, and are then stuck in `InProgress` for 2+ minutes until the stale-rescue fires. The aircraft enters them before the rescue runs → FUSE misses → on-demand network fetch.

## Fix plan (narrow, TDD)

Four parts, one PR. Scope ~5-8 files, ~400 lines net including tests.

### 1. Defer `mark_in_progress` until after successful submission

Move the `boundary_strategy.mark_in_progress()` call from before `execute_plan()` to after it returns. Only mark regions for which **all** of their planned tiles were accepted by the executor (submitted, not throttle-overflowed or channel-rejected). Regions with partial or zero submission remain unmarked and naturally re-enter `new_regions_with_extent` on the next cycle.

### 2. Carry the explicit submitted-tile set through `ExecutionResult`

Change `ExecutionResult.submitted: usize` to `submitted_tiles: Vec<TileCoord>`, populated by `execute_plan` as tiles are accepted by the executor. Update Part 1's per-region marking check to use a **positive** criterion (`all planned tiles ∈ submitted_tiles`) instead of a **negative** one (`no planned tile ∈ pending`).

**Why this matters beyond Part 1:** `plan_executor` caps pending at `MAX_PENDING_TILES = 2000` (plan_executor.rs:165-170). When a first-cycle plan has 5,000 tiles and the throttle admits only 19, pending ends up with 4,981 tiles — truncated to 2,000. **2,981 tiles are silently dropped** (not submitted, not retained as pending). Under Part 1's negative check, those dropped tiles wrongly count as "submitted" because they're not in `pending`. A region whose tiles fell entirely into the dropped range would be incorrectly marked `InProgress` with zero tiles actually submitted — reproducing the original bug in a narrower form.

With the positive check, dropped tiles correctly count as NOT submitted → their regions correctly stay unmarked → they naturally re-enter `new_regions_with_extent` on the next cycle for retry.

Additional benefit: `submitted_tiles.len()` gives an exact per-cycle count for Part 4's `tiles_planned / submitted / pending` INFO log, without a separate counter.

*Correction note:* An earlier draft of this fix proposed "merge pending_tiles instead of replacing" (extend + truncate in core.rs line 664). Closer reading showed the replace was safe — the pending-drain path uses `std::mem::take` before calling `execute()`, so the field is always empty at the moment of assignment. That scope is discarded; this Part 2 is the real fix.

### 3. Back `promote_completed_regions` with the authoritative DDS disk cache

Replace the `cached_tiles: &HashSet<TileCoord>` parameter with `dds_disk_checker: &Arc<dyn DdsDiskCacheChecker>`. For each `InProgress` region:
- Compute its tile set via `tiles_for_region` (existing)
- **Check-all tiles with short-circuit on first miss** against `DdsDiskCacheChecker`
- If all present, promote to `Prefetched`

Add a `tile_exists_blocking` sync method to the `DdsDiskCacheChecker` trait. The underlying DashMap lookup is already sync; the async wrapper is gratuitous. This lets us drop the `block_in_place` + `block_on` hack that `evaluate_stale_regions` currently uses, simplifying both callsites.

The `cached_tiles` HashSet is **not** removed in this fix. Its only remaining role is a fast-path optimisation inside `filter_memory_cache`; it's no longer used as an authoritative cache tracker. Deletion is deferred to the v0.5.0 CacheLocator refactor.

### 4. Instrumentation

Surface the bug class in telemetry so recurrence is immediately visible:
- **Per-cycle INFO log**: `tiles_planned=N tiles_submitted=M tiles_pending=K strategy=…` — persistent `planned ≫ submitted` = this bug
- **Per-maintenance-cycle INFO log**: `regions_in_progress=N regions_prefetched=M regions_nocoverage=K` — state-machine stall visibility
- **TUI metric**: cumulative `planned / submitted` ratio. Deviation from ~1.0 over time surfaces the pattern

## Verification plan

After implementation, re-run the LOWW westbound test with `xearthlayer run --debug` and confirm:
- `Stale InProgress region promoted — tiles found on DDS disk` count drops from 61 (9h log) to near 0
- `Promoted InProgress regions to Prefetched` becomes the dominant promotion path (previously 4 of 65)
- The departure-bubble DSFs (+47..+49 / +015..+018) show no FUSE misses on second flight from same airport
- Per-cycle `tiles_planned` vs `tiles_submitted` ratio stabilises near 1.0 after the transition ramp

## Known limitation (to be tracked separately)

Promotion correctness is downstream of **scenery index completeness**. If a `.ter` file fails to parse or a package isn't fully scanned, `tiles_for_region` under-reports the tile set and `promote_completed_regions` can fire prematurely. This is a pre-existing risk, not introduced by this fix, but worth tracking. Follow-up issue to be filed.

## Architectural follow-up — v0.5.0 CacheLocator

The shadow-state drift pattern this bug exposes is a class defect, not a one-off. The narrow fix above addresses the specific failure mode without removing the shadow. A separate v0.5.0 tracking issue covers the full architectural cleanup: eliminate `cached_tiles`, simplify `PrefetchedRegion` states, remove retention machinery, introduce a unified `CacheLocator` as the single query interface so cache internals don't leak into prefetch / FUSE / TUI code.

## Historical context

Earlier comments on this issue preserve the full diagnostic trail: Phase 1 scope-narrowing hypothesis, Phase 2 log-data refutation of the initial framing, and the three verification queries that pinpointed the mark-before-submit mechanism.

Recorded from: LOWW westbound test flight, 2026-04-11, debug-telemetry branch + PR #174 instrumentation.

samsoir:
## Summary

Fixes #172 — the root cause of "tiles not ready on approach" bugs on long-haul flights, plus the post-flight follow-ups that fell out of in-sim testing. Six related defects now addressed in one cohesive branch; the full `CacheLocator` SSOT refactor (#175) is still deferred to v0.5.0.

## Background

Verified against a 9-hour LOWW westbound debug log (2.8 GB). Phase 2 investigation found that **4,828 FUSE-missed tiles clustered in 12 contiguous DSFs** forming the departure bubble (+47..+49 / +015..+018). The prefetch coordinator's `get_tiles_for_region` was called for all 12 regions and computed hundreds of tiles each — but almost none reached the executor. Root cause: regions were committed to `InProgress` state **before** the submission result was known, and then filtered out of `new_regions_with_extent` forever.

Signature anti-pattern: **61 "stale rescue" promotions vs. only 4 normal-path promotions** in the 9-hour log. The shadow-state `cached_tiles` HashSet was failing to track ~94% of actually-cached tiles; the rescue path was carrying the system.

### Part 1 — Defer `mark_in_progress` until after successful submission

`mark_in_progress` moves from the cruise branch of `update()` to `execute()`, called only for regions where **every** planned tile appears in the authoritative submitted set. Regions with deferred / channel-rejected / throttle-overflowed tiles stay unmarked and naturally re-enter the new-regions scan on the next cycle.

New field: `current_plan_regions: HashMap<TileCoord, DsfRegion>` — transient per-plan tile→region mapping, populated by `update()`, consumed by `execute()`, cleared each cycle.

### Part 2 — `ExecutionResult` carries the explicit submitted-tile set

Changed `ExecutionResult.submitted: usize` → `submitted_tiles: Vec<TileCoord>`. Part 1's completion check uses a **positive** criterion (`∈ submitted_tiles`) instead of a negative one (`∉ pending`). Corrects the `MAX_PENDING_TILES` truncation edge case: tiles dropped by the 2000-tile cap are in **neither** list, so a negative check would wrongly classify them as submitted. The positive check correctly leaves their regions unmarked.

### Part 3 — `promote_completed_regions` queries the DDS disk cache directly

Signature change: `cached_tiles: &HashSet<TileCoord>` → `dds_disk_checker: Option<&Arc<dyn DdsDiskCacheChecker>>`. For each `InProgress` region, check-all tiles with short-circuit on first miss against the authoritative DDS disk cache.

Added `tile_exists_blocking` sync method to the `DdsDiskCacheChecker` trait. `DdsDiskCacheBridge` implements it via `block_in_place` internally — encapsulating the hack that was previously inline in `evaluate_stale_regions`.

### Part 4 — Cycle-level instrumentation

- **Per-cycle INFO**: `Prefetch cycle summary strategy=... tiles_planned=N tiles_submitted=M tiles_pending=K tiles_dropped=D deferred=B`.
- **Per-maintenance-cycle INFO**: `Region maintenance: state distribution regions_in_progress=N regions_prefetched=M regions_nocoverage=K`.

### DDS disk cache key mismatch — the actual root cause of drain-cycle disk spam

Parts 1-4 fixed the lifecycle but flight-testing revealed regions still weren't being promoted. `filter_dds_disk_cache` and `promote_completed_regions` were passing `chunk_origin()` triples to the disk cache checker, but the DDS disk cache is keyed by tile coords (`tile:{zoom}:{row}:{col}`). Every lookup silently returned `false`, so previously-prefetched tiles were re-submitted each cycle and served as "cache hits" via a 10MB disk read — manifesting as continuous disk activity during what should have been quiet drain cycles.

Added a regression test that populates the checker with chunk coords and asserts zero filter hits, so the two coordinate systems can't be conflated again.

`DiskCacheProvider::contains()` now trusts the LRU index instead of `tokio::fs::metadata()`-verifying per call. Stale entries are cleaned up reactively in `get()` on NotFound — moving the cleanup cost from "every contains call, always" to "only when a get actually discovers the missing file."

### Unify ground/cruise prefetch under a single `PrefetchBox`

The old ring-based `GroundStrategy` produced only 18 DSF tiles — inside X-Plane's ~3° scenery window — and relied on dead `scene_tracker.loaded_bounds()` wiring (never called in production). Since the Web API always provides telemetry, the ring's heuristic edge detection is unnecessary.

- Deleted `ground_strategy.rs` (746 lines).
- Ground phase now uses `PrefetchBox` with symmetric bias (0.5); cruise keeps the heading-biased 0.8 variant. Retention updates remain cruise-only (updating on the ground would evict legitimate prewarm state).
- Introduced `BoxShape { extent, max_bias }` so the coordinator passes both per-call without mutating the box and without tripping `clippy::too_many_arguments`.
- Removed deprecated keys (added to `DEPRECATED_KEYS`): `prefetch.max_tiles_per_cycle`, `prefetch.default_window_rows`, `prefetch.window_lon_extent`, `prefetch.ground_ring_radius`.
- Unified extent config: single `box_extent` (default 7.0°) used as ground-fixed extent and cruise maximum.

### Debug map now matches reality (single source of truth)

The map's box overlay was rebuilt from default config on every refresh (always 0.8 bias) — so the ground-phase overlay rendered a heading-biased box even when the coordinator was using symmetric. The region colors were also painting green the moment any single tile completed.

- Coordinator computes bounds once per cycle and publishes them to `SharedPrefetchStatus` as `BoxBoundsSnapshot`. The debug map reads them verbatim. No recomputation, no drift.
- `collect_regions` now treats the GeoIndex as authoritative for lifecycle state. A region stays yellow `InProgress` until the completion promoter verifies every tile is on disk, then flips green `Prefetched`. Activity counters no longer win.
- New orange `Mixed` upgrade: when a region is `Prefetched` AND FUSE has touched it (hits or misses), the color changes to orange to visualise "prefetch successfully served X-Plane's demand."

## Test plan

- [x] `make pre-commit` passes (fmt + clippy -D warnings + 2,378 tests, 0 failures)
- [x] Part 1-4 unit tests: 7 new + 4 rewritten
- [x] Post-flight follow-up tests: regression guard on DDS cache key usage, SSOT conversion for `BoxBoundsSnapshot`, classification matrix for region state
- [x] **Flight-tested at LOWW**:
  - Ground prefetch converges cleanly (symmetric box, all 49 DSF regions queued and drained)
  - Takeoff transitions box to heading-biased cruise
  - Boundary crossings served entirely from DDS disk cache — **zero network I/O** on re-visited terrain
  - Debug map overlay visibly tracks phase bias (symmetric on ground, elongated forward in cruise)

## Stats

- **Total**: 28 files changed, ~1,200 lines of net deletion (mainly the 746-line ground_strategy + deprecated config paths)
- **Known limitation** tracked as #176 (scenery index completeness affects promotion correctness — pre-existing)
- **Architectural follow-up** tracked as #175 (unified `CacheLocator` SSOT for v0.5.0)


samsoir:
## Flight test surfaced a regression — hotfix pushed (31c1273)

A live test today revealed that the Parts 1-4 fix in this PR was **strictly worse** than pre-fix behaviour under realistic conditions. Post-climb-out, the coordinator produced zero prefetch submissions for ~10 minutes of cruise, culminating in 20-second hard sim freezes at scenery-window boundary crossings.

### Smoking gun

```
Prefetch plan filter pipeline summary
  raw_plan_tiles=200 cache_skipped=200 patch_skipped=0 disk_skipped=0 remaining=0
```

Every plan: 200 raw tiles → 200 "cached" → 0 remaining → execute never called → no cycle summary → nothing marked InProgress → nothing to rescue.

### Root cause

`filter_memory_cache` was consulting the `cached_tiles` HashSet shadow as a fast-path before querying the authoritative memory cache. Because the memory cache is intentionally small (request absorber, not working-set holder), entries get evicted from moka within seconds — but the shadow keeps saying "cached." The filter reported false cache hits, emptying every plan.

This is **exactly the Phase 1 hypothesis** from the original #172 body (`cached_tiles` ghost entries). I moved away from it when the Phase 2 data suggested mark-before-submit was the primary defect — but both bugs were real and compounding. Parts 1-4 fixed the mark-before-submit defect; the shadow staleness needed its own fix.

### Why it got worse after Parts 1-4

- **Before my fix**: stale shadow → filter lies → but regions still get pre-submission marked InProgress → `evaluate_stale_regions` rescue path could eventually promote them when their tiles happened to land on disk (61 rescue events / 9h in the bug flight).
- **After my fix alone**: stale shadow → filter lies → plans empty → Part 1's positive-check marking never fires → regions stay untracked → rescue path has nothing to rescue. Silent starvation.

### My scope-call error

In the PR description I wrote that `cached_tiles` would be left as "a fast-path optimisation" with deletion "deferred to v0.5.0 CacheLocator." That was wrong. The shadow was load-bearing for correctness, not a speedup. I violated my own `feedback_ssot_no_compromises` memory in the process.

### Hotfix (commit 31c1273)

- Remove the `cached_tiles.contains(tile)` fast-path in `filter_memory_cache`. Always query the authoritative memory cache.
- Stop writing to the shadow on hit (decoupled from filter).
- Retain the parameter for compatibility with the retention bookkeeping (`evict_cached_tiles_outside_retained`); full removal is v0.5.0 (#175).

### 4 new regression tests (catch the bug in <5s)

1. `test_filter_does_not_trust_stale_shadow_over_authoritative_cache` — direct regression guard at the filter level
2. `test_filter_does_not_write_to_shadow` — ensures the filter no longer populates the stale structure
3. `test_filter_correctly_splits_partial_real_cache` — proof the filter queries the real cache via an even-rows-hit mock
4. `test_coordinator_submits_work_despite_stale_shadow_and_empty_cache` — **end-to-end integration test** requested by the user: 5000-entry pre-populated shadow + always-miss memory cache + full coordinator flow, asserts `submitted > 0`

Baseline 2378 → 2382 tests. `make pre-commit` passes.

### Status

- Flight test aborted mid-flight because of the regression
- Hotfix needs a second flight test to verify
- The original Parts 1-4 issues are still fixed; the shadow-staleness fix is additive

mmaechtel:
Watching this with great interest :) Happy to help test whenever useful — though I should mention I'm not really a long-haul flyer, so my runs will be on the shorter side.

samsoir:
@mmaechtel definitely do not test yet :-) working through bugs on bugs, but making progress. All of the throttling on executor in the past has overly complicated the prefetch system to the point of preventing it from doing its job. Given the other improvements to execution efficiency I am removing a number of the throttle controls and relying on the job queue to govern throughput - which was always the intent.

The issue that originally presented itself was that prefetch would only cache "some" of the tiles needed, not all of them. And down the rabbit hole I went :sweat:

mmaechtel:
Loving v0.4.3 — performance on short flights is so good I've started disabling all my self-made zOrthos and going full streaming. Even experimenting with higher-resolution meshes on top (still tweaking that part). So hearing that caching is about to get even better is great news 🙂

Totally understood on holding off testing — I'll wait for your green light.

samsoir:
Great to hear! When this fix is ready, I will cut 0.4.4 immediately as this is priority. The rest of the 0.4.4 items will push to a later 0.4.x release.

samsoir:
@mmaechtel local testing with this latest change is working well for me. X-Plane is loading all perfetched tiles from disk. I need to tune the performance a little to leave more room for X-Plane, but the boundary transitions are largely not noticeable from the flight deck now. feel free to test.

Still some refinement to do but likely this will ship as v0.4.4

samsoir:
:shipit:

Issue #182, opened 2026-05-05
Diagnostics: provider performance probe command


samsoir:
## Background

Users have no easy way to compare provider performance from their own network location before committing to a provider in their config. Anecdotal user reports (e.g. Google Maps faster than Bing in some regions) suggest meaningful variance, but there's no tool to verify this empirically.

A simple probe command, sampling the currently-configured provider plus all *free* providers, would let users make a data-driven choice. This pairs with the per-provider failure-rate telemetry (companion issue) — telemetry tells you "your current provider is degraded right now," the probe tells you "here's what to switch to."

## Proposal

A new diagnostics subcommand:

```
xearthlayer diagnostics probe [--airport <icao>] [--zoom <zoom>] [--samples <n>]
```

For each provider in scope, download a sample of tiles and report timing and success rate.

### Provider scope (default)

- Currently configured provider (whatever's in `[provider]` config)
- All free providers (`bing`, `go2`)

Excluded by default: `google` (paid, requires API key — would surprise users with billable requests).

A future `--include-paid` flag could opt in to paid providers if a user wants to compare. Out of scope for this issue.

### Sampling

- Default lat/lon: aircraft position if Web API connected, else a hardcoded reference (e.g. Denver Int, Colorado KDEN)
- Default zoom: 16 (chunk zoom) or 12 (tile zoom) — TBD which is more representative
- Default samples: e.g. 10 chunks per provider, downloaded sequentially with timing measured per chunk, and burst of 10 simultaneously to simulate real world performance

### Output

Tabular, sortable by latency:

```
Provider   Samples  Success  P50 latency  P95 latency  Mean bytes
google     10       100%     142ms        287ms        18.4 KB
bing       10       100%     203ms        412ms        21.1 KB
go2        10       90%      178ms        624ms        17.8 KB
```

Plus a one-line recommendation: "Fastest free provider for your location: google" (or similar).

## Implementation notes

- Reuse existing `AsyncBingMapsProvider` / `AsyncGo2Provider` / `AsyncGoogleMapsProvider` clients.
- Probe should not pollute caches (chunk disk or otherwise) — bypass cache writes for these requests, or use a separate temp scope.
- Honor existing HTTP semaphore limits; don't blast 30 simultaneous chunks at the provider.
- Log structured output to file in addition to TUI table, so users can attach to bug reports.

## Scope

- One-shot diagnostic command, not a continuous monitor (telemetry covers continuous monitoring).
[...]

Issue #181, opened 2026-05-05
Per-provider failure-rate telemetry + TUI surface


samsoir:
## Background

Discussion thread on #180 highlighted that users have no current visibility into *why* tile downloads are failing. Network outage, cloud-provider downtime, rate limiting, and auth rejection all surface to the user as "magenta tiles," with no way to distinguish them or decide whether switching providers would help.

Auto-fallback between providers has real design cost (visual seams, recovery policy, mode-#3-vs-#1 detection ambiguity) and isn't worth building until we have data showing users actually hit provider-side failures often enough to justify it. Observability comes first.

## Proposal

Track and surface per-provider failure rates in real time, so users can make informed manual decisions about provider switching.

### Metrics

For each configured provider, track over a sliding window (suggested: last 60s):

- Total chunk requests
- Successful responses
- Categorized failures:
  - `RateLimited` (HTTP 429, plus provider-specific 403/503 patterns)
  - `AuthRejected` (HTTP 401/403 with provider-specific signaling)
  - `Timeout` (request_timeout exceeded)
  - `NetworkError` (DNS, connection refused, TLS, generic transport)
  - `OtherHttp` (any other non-2xx)

Categorization matters because *RateLimited* and *AuthRejected* are strong signals to switch provider, while *Timeout* and *NetworkError* are ambiguous (could be user-side).

### TUI surface

Add a compact indicator to the cache widget area or as its own status line:

```
Provider: bing  errors: 0.3% (60s)  cat: 0 rl, 0 auth, 1 net
```

Color/state thresholds (suggested):
- Green: <1% error rate
- Yellow: 1-10% error rate
- Red: >10% error rate, or any *RateLimited* / *AuthRejected* in window

### Diagnostics output

`xearthlayer diagnostics` gains a section showing recent provider error breakdown, sourced from the same metrics.

## Scope

- Metrics tracking only — no automatic provider switching, no fallback, no decision-making.
- Single configured provider (current model). Multi-provider tracking can come later if/when fallback ships.
- Sliding window in memory; no persistence across runs.

## Out of scope

- Automatic provider fallback (deferred behind real data from this telemetry).
- Per-region or per-zoom failure breakdown.
- Cross-session error history.

## Validation

- Manual test: simulate 429 by shaping traffic / blocking specific URLs; confirm category and threshold trip correctly.
- Verify the TUI indicator updates in real time during a normal flight (should sit at green / 0%).
[...]

PR #183, opened 2026-05-06
fix(cache): never persist tiles with failed chunks (#180)

### LINKED ISSUE #180: Question: how should the cache handle tiles built from failed downloads?
Hi — 0.4.4 works great here, thanks. One observation I went looking into and want to check with you before filing as a bug, because I'm probably misreading something.

## What I saw

A ~10 minute internet outage mid-flight. Tiles X-Plane requested during the outage came up magenta — fine. What surprised me: after the connection came back, those tiles stayed magenta, and `Reload Scenery` from X-Plane's Developer menu didn't recover them. With a stable connection I never see magenta tiles, and during intermittent hiccups they seem to "go away" during flight — though from reading the code I now suspect that's higher-LOD tiles painted over them, not the cache being repaired.

## What I think I'm seeing in the code (could be wrong)

- `DownloadChunksTask` (`tasks/download_chunks.rs:151-161`) returns `SuccessWithOutput` regardless of how many of the 256 chunks failed. `success_count` / `failure_count` only feed a `debug!` log — I couldn't find anywhere they gate downstream behavior.
- `assemble_chunks` (`tasks/build_and_cache_dds.rs:322-356`) always returns `Ok(canvas)`. Two paths fill magenta: failed downloads (`:350`) and failed JPEG decodes of successful downloads (`:345`).
- The cache-write `tokio::spawn` blocks (`:259-268` memory, `:278-298` DDS disk) appear unconditional — no failure-ratio check.
- From the executor (`jobs/dds_generate.rs:191-198`, `executor/daemon.rs:850-860`), the job counts as `Succeeded` because no task returned `Failed`. The "return empty `Vec`, don't cache" path on `:850-860` only seems to fire on task panics or encoder errors.
- The DDS validator (`fuse/placeholder.rs:159-191`) checks empty / expected size / DDS magic. A magenta 4096×4096 BC1 DDS passes all three, and it runs on the path *out* to X-Plane (`fuse3/shared.rs:424`), downstream of the cache writes.
- `DiskCacheProvider::get` (`cache/providers/disk.rs:318-336`) returns whatever bytes are on disk — no mtime / freshness / integrity check.
- No TTL on either tier (`cache/config.rs:208`, no TTL field on `DiskProviderConfig`). Disk GC fires only above ~95% of budget (`cache/gc_scheduler.rs:46`). No scrub/repair daemon I could find.

If that's right, a tile mostly assembled from magenta would be structurally indistinguishable from a successful one and persist on disk indefinitely. With my config (`disk_size = 1500 GB`, `dds_disk_ratio = 0.6`) GC effectively never fires, which would make this more visible than on defaults.

## Question

Is "no gate on failure_count" intentional, or did I miss a control path?

## Environment

- v0.4.4 (commit `507ac9c`), Bing Maps, Linux
- `disk_size = 1500 GB`, `dds_disk_ratio = 0.6`

samsoir:
## Summary

- Gates memory + DDS-disk cache writes in `BuildAndCacheDdsTask` on `ChunkResults::is_complete()`.
- Tiles with any failed chunks are still served to X-Plane (so the sim doesn't stall during outages) but never persisted to memory or DDS disk.
- Chunk-disk tier was already correct (only successful HTTP downloads cached); it now functions as a natural retry-resume buffer on re-request — only the chunks that previously failed get re-fetched.
- Documentation: per-tier completeness rules added to CLAUDE.md cache section; vestigial "80% minimum success rate" line removed (described the pre-task-framework orchestrator and didn't gate anything).
- Tests: two new task-level tests covering both branches (incomplete tile → no cache writes; complete tile → both cache writes fire).

## Background

Reported by @mmaechtel in #180. After a ~10 minute internet outage mid-flight, magenta tiles persisted in his cache and didn't recover even after `Reload Scenery`. Investigation confirmed the bug: `DownloadChunksTask` returned `SuccessWithOutput` regardless of failure count, `assemble_chunks` always returned `Ok(canvas)` with magenta filled in for failed chunks, and the two `tokio::spawn` cache writes in `BuildAndCacheDdsTask` were unconditional. The validator at `fuse/placeholder.rs` checks size/magic, both of which a magenta BC1 tile satisfies, so poisoned tiles passed all downstream guards and persisted indefinitely (Michael's `disk_size = 1500 GB` config made GC-driven natural eviction effectively impossible).

## Per-tier policy (now reflected in CLAUDE.md)

| Tier | Granularity | Rule |
|------|-------------|------|
| Chunk disk | per-chunk | Cache only successful HTTP downloads. *(Already correct.)* |
| DDS disk | per-tile | Cache only when `ChunkResults::is_complete()`. |
| Memory | per-tile | Same. |

## Test plan

- [x] `make pre-commit` passes locally (fmt + clippy `-D warnings` + full unit/integration/doc tests)
- [x] New test `incomplete_tile_does_not_write_to_caches` fails on unmodified main, passes after the fix (TDD red→green confirmed)
- [x] New test `complete_tile_writes_to_both_caches` passes (regression guard for the happy path)
- [x] Local flight test by maintainer (network outage simulation, confirmed magenta tiles re-fetch on reconnect rather than serving cached magenta)
- [ ] CI green

## User-facing repair

Users who saw magenta tiles persist across an outage on v0.4.4 or earlier should run:

```
xearthlayer cache clear
```

This is documented in the CHANGELOG entry under `[Unreleased]` and called out in the comment thread on #180.

## Related

- Fixes #180
- Discussion thread on #180 also produced #181 (per-provider failure-rate telemetry) and #182 (provider performance probe diagnostic), both targeting v0.5.0. Those are separate observability work, not part of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Tiles assembled from failed chunk downloads are no longer cached** ([#180](https://github.com/samsoir/xearthlayer/issues/180)): Network failures during tile generation produced magenta-filled DDS tiles that were structurally valid (correct size + DDS magic) and persisted indefinitely in the memory and DDS disk caches, surviving even X-Plane's "Reload Scenery". On reconnect, re-requests would short-circuit to the poisoned cache instead of re-fetching. Cache writes are now gated on `ChunkResults::is_complete()`: tiles with any failed chunks are still served to X-Plane (so the sim doesn't stall during outages) but never persisted to memory or DDS disk. The chunk-disk tier was already correct (only successful HTTP downloads were cached), so it now functions as a natural retry-resume buffer on re-request — only the chunks that previously failed get re-fetched. Users who saw magenta tiles persist across an outage on v0.4.4 or earlier should run `xearthlayer cache clear` to purge any poisoned tiles.
+
 ## [0.4.4] - 2026-04-16
 
 ### Changed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,7 +40,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 3. **Download Orchestrator** (`xearthlayer/src/orchestrator/`)
    - Parallel download of 256 chunks (16×16) per tile
    - Configurable concurrency, timeout, and retry logic
-   - 80% minimum success rate requirement
+   - Failed chunks are not cached (retry on next request); see Cache System for tile-level rules
 
 4. **DDS Compression** (`xearthlayer/src/dds/`)
    - BC1/BC3 (DXT1/DXT5) compression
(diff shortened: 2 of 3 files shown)


samsoir:
:shipit:

Issue #186, opened 2026-05-08
[Bug]: package download errors


dee-sea:
### Bug Description

Hello and thanks a lot for making this project open-source.

I'm quite new to xplane but a longtime linux user and gamer. I just been aware of this project through
your YT video and I had to try it.  I face a problem downloading the eu package. I mostly fly in europe and was able to download eu2 without errors but when I try to get the eu package i always  get download errors:

```
 zzXEL_eu_ortho-0.1.1.tar.gz.aa (failed: failed to download https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.aa: Read error: request or response body error)
  zzXEL_eu_ortho-0.1.1.tar.gz.ab (failed: failed to download https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.ab: Read error: request or response body error)
  zzXEL_eu_ortho-0.1.1.tar.gz.ac (failed: failed to download https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.ac: Read error: request or response body error)
  zzXEL_eu_ortho-0.1.1.tar.gz.ae (failed: failed to download https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.ae: Read error: request or response body error)
  zzXEL_eu_ortho-0.1.1.tar.gz.ao (failed: failed to download https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.ao: error sending request for url (https://github.com/samsoir/xearthlayer-regional-scenery/releases/download/eu-v0.1.1/zzXEL_eu_ortho-0.1.1.tar.gz.ao))
[...]

PR #190, opened 2026-05-09
fix(diagnostics, packages): disk-space awareness (#188)

### LINKED ISSUE #188: Disk-space awareness: diagnostics shows wrong filesystems, no pre-install space check
## Summary

Three related defects that combine to give users (especially on immutable / atomic Linux distros and small-disk installs) misleading diagnostic output and no pre-flight protection against disk-full failures during package installs.

Surfaced while investigating [#186](https://github.com/samsoir/xearthlayer/issues/186) (dee-sea on Bazzite). Their diagnostics reported `Root filesystem: 45M used of 45M (100%)` and `Cache directory: (not created)` — both misleading or wrong. Their actual problem was network-side (separate investigation), but the diagnostics output offered no useful disk-side signal because it wasn't measuring the things that actually matter to XEL.

## A — Diagnostics reports the wrong filesystem on immutable OSes

`xearthlayer/src/diagnostics/report.rs:228` runs `df -h /` to report root filesystem usage:

```rust
if let Ok(output) = Command::new("df").args(["-h", "/"]).output() {
    // ...
    info.root_total = Some(parts[1].to_string());
    info.root_used = Some(parts[2].to_string());
    info.root_percent = Some(parts[4].to_string());
```

On Bazzite, Silverblue, Kinoite, SteamOS, and other ostree-based atomic distros, the rootfs is a deployed OSTree image sized exactly to its content. `100%` is the **steady state**, not a problem. Reporting it alarms users for no reason and obscures the filesystems that actually matter to XEL (cache directory, packages install location).

### Proposed fix

1. Detect immutable / atomic OS context via `/run/ostree-booted` existence or `os-release` (`VARIANT_ID` containing `atomic`, `silverblue`, `kinoite`, etc.).
2. Replace single `df /` with `df` on each of: `cache.directory`, `packages.install_location`, and `xplane.scenery_dir` (the writable paths XEL actually uses).
3. On immutable OSes, either skip the rootfs line entirely or annotate it: `Root filesystem: 45M / 45M (100%) — normal on atomic distro`.

## B — Diagnostics checks the wrong cache path

`xearthlayer/src/diagnostics/report.rs:244`:

```rust
let cache_dir = home.join(".cache/xearthlayer");
```

But the default `cache.directory` config value is `~/.xearthlayer/cache` (note: under the config dir, not under `~/.cache/`). The diagnostics will always show `Cache directory: (not created)` for users on the default config, even when they have a fully populated cache.

### Proposed fix

Read `cache.directory` from the loaded config and `du -sh` against that path. Fall back to the documented default (`~/.xearthlayer/cache`) if no config is loaded. If the configured path is `~`-prefixed, expand it before checking.

## C — No pre-flight disk-space check before package install

A grep for `available_space`, `statvfs`, etc. across the codebase finds nothing. `recommended_disk_cache()` returns a hardcoded `40 * GB` regardless of free space. `xearthlayer packages install eu` happily downloads multi-GB archives without ever checking that the install location has space for them — failures surface only mid-stream as I/O errors when writes start failing.

### Proposed fix

Before kicking off `MultiPartDownloader::download_all()` in `PackageInstaller::install_from_metadata()`:

1. Sum `part_size_bytes` across all parts in the metadata.
2. Add a buffer for tar extraction (decompressed size is roughly 2x compressed for satellite imagery).
3. `statvfs(install_location)` to get available bytes.
4. If `available < required`, fail-fast with a clear error:
   ```
   Insufficient disk space at <install_location>:
     required: 8.5 GB (4.2 GB download + 4.3 GB extracted)
     available: 2.1 GB
   ```

Use the `nix::sys::statvfs` crate (already in the dependency tree if present, otherwise small addition) for portable POSIX free-space query.

## Why one issue, not three

A, B, and C share a common helper: \"what filesystem does XEL actually use, and how much space is on it?\" Fixing them together means writing that helper once. They also have a shared user impact (the diagnostics-and-install-disk-space story), and a shared test surface (mock `statvfs` and a stub config).

## Validation

- Manual on Bazzite or Silverblue VM: confirm diagnostics no longer alarms on rootfs at 100% and reports cache/install paths instead.
- Manual on a small-disk install (`/tmp` mount with limited space) configured as `packages.install_location`: confirm `packages install eu` fails fast with a clear message rather than mid-stream I/O error.
- Default-config user (no config edits): confirm `xearthlayer diagnostics` reports the actual cache path (`~/.xearthlayer/cache`) and its size, not `(not created)` for an existing populated cache.

## Out of scope

- Capping `recommended_disk_cache()` to a fraction of actual free space. Sensible enhancement but not required for this fix; track separately if ever wanted.
- Real-time disk-space monitoring during runtime cache writes (separate concern; cache GC handles eviction once budget is exceeded).

## Related

- #186 — user report on Bazzite where this surfaced
- #187 — separate defect (package install hang) on a different code path; not superseded by this issue

samsoir:
## Summary

Three related defects on immutable-OS and small-disk installs, fixed together because they share a helper module for filesystem capacity queries. Surfaced while investigating [#186](https://github.com/samsoir/xearthlayer/issues/186) (dee-sea on Bazzite reported `Root filesystem: 100%` and `Cache directory: (not created)` even though both were normal/wrong respectively).

### A. Diagnostics reports paths that matter

- `DiskInfo` gains `cache_dir_path`, `cache_dir_available`, `install_location_path`, `install_location_available`, `is_immutable_os`.
- Output now leads with cache directory (used + available) and packages location (available), then renders rootfs annotated `— normal on atomic distro` when `/run/ostree-booted` exists.

Before / after for a Bazzite user:
```
Root filesystem: 45M used of 45M (100%)
Cache directory: (not created)
```
becomes:
```
Cache directory: ~/.xearthlayer/cache (12.3 GB used, 28.7 GB available)
Packages location: ~/.xearthlayer/packages (28.7 GB available)
Root filesystem: 45M used of 45M (100%) — normal on atomic distro
```

### B. Diagnostics reads cache path from config

`DiskInfo::collect()` now loads `ConfigFile` and calls `from_config()`, sourcing `cache.directory` from the config rather than hardcoding `~/.cache/xearthlayer`. Default-config users (cache at `~/.xearthlayer/cache`) finally see the correct path.

### C. Pre-flight disk-space check before package install

New module `manager::disk_check` with:
- `FsInfoProvider` trait (real impl delegates to `system::fs_info`; tests inject low-available stubs)
- `check_disk_space(provider, path, required) -> ManagerResult<()>`
- `SPACE_BUFFER_MULTIPLIER = 2` (compressed + extracted coexist briefly)
- `ManagerError::InsufficientDiskSpace { path, required_bytes, available_bytes }`

Inserted in `PackageInstaller::install_from_metadata` after `query_sizes()` populates expected totals and before `download_all()` kicks off network work. Failures look like:

```
insufficient disk space at /home/user/.xearthlayer/tmp/install_eu_ortho_0.1.1: required 8.5 GB, available 2.1 GB
```

## Shared helper

New module `system::filesystem`:
- `FilesystemInfo { total_bytes, available_bytes }`
- `fs_info(&Path)` via `nix::sys::statvfs` (nix v0.31, already in workspace tree, promoted to direct dep)
- `is_immutable_os()` checks `/run/ostree-booted` (the standard ostree marker)

## Test plan
- [x] Phase 1: 3 tests in `system::filesystem` (capacity for /tmp, error for nonexistent path, immutable-OS detection doesn't panic)
- [x] Phase 2: 5 tests in `diagnostics::report` (config-driven cache path, available bytes for existing paths, ancestor walk for nonexistent paths, install_location fallback, ancestor-walk helper)
- [x] Phase 3: 5 tests in `manager::disk_check` (sufficient space, exact-equal boundary, insufficient space, provider error path, error Display format)
- [x] `make pre-commit` passes locally (full 2394-test suite + clippy + fmt)
- [x] CI green
- [x] Manual diagnostics check on a Bazzite VM (or any ostree-booted system) — out-of-scope for CI; will document any post-merge findings

## Out of scope (per issue body)

- Capping `recommended_disk_cache()` to actual free space (separate concern, defer)
- Real-time disk-space monitoring during runtime (cache GC handles eviction once budget exceeded)
- Windows/macOS support (Linux-only project)

Fixes #188

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Disk-space awareness for diagnostics and package installs** ([#188](https://github.com/samsoir/xearthlayer/issues/188)): Three related defects on immutable-OS and small-disk installs:
+  - Diagnostics now reports cache directory and packages location with available bytes alongside used (where it actually matters), instead of only `df /` of the rootfs (which on Bazzite, Silverblue, Kinoite, SteamOS and other ostree-based atomic distros is intentionally sized to its content and reads `100%` as the steady state). The rootfs line is now annotated `— normal on atomic distro` when XEL detects an ostree-deployed root via `/run/ostree-booted`.
+  - Diagnostics now reads `cache.directory` from config rather than checking a hardcoded `~/.cache/xearthlayer` (which never matched the default `~/.xearthlayer/cache`), so the "Cache directory: (not created)" line was wrong for default-config users with a populated cache.
+  - Package installs (`xearthlayer packages install <region>`) now run a pre-flight disk-space check after sizing parts via HEAD requests but before any bytes hit the network. Insufficient space fails fast with `insufficient disk space at <path>: required X GB, available Y GB` instead of mid-stream I/O errors during the download.
+
 ## [0.4.5] - 2026-05-05
 
 ### Fixed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4775,6 +4775,7 @@ dependencies = [
  "intel_tex_2",
  "libc",
  "moka",
+ "nix",
  "parking_lot",
  "pollster",
  "proptest",
--- a/xearthlayer/Cargo.toml
+++ b/xearthlayer/Cargo.toml
@@ -50,6 +50,7 @@ futures-util = "0.3"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio-tungstenite = { version = "0.24", features = ["connect"] }
+nix = { version = "0.31", default-features = false, features = ["fs"] }
 tracing-chrome = { version = "0.7", optional = true }
 axum = { version = "0.7", optional = true }
 
--- a/xearthlayer/src/diagnostics/report.rs
+++ b/xearthlayer/src/diagnostics/report.rs
@@ -2,6 +2,7 @@
 
 use std::fmt;
 use std::fs;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 /// A comprehensive system diagnostics report.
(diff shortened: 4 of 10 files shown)


samsoir:
:shipit:

PR #191, opened 2026-05-09
fix(packages): first-failure abort with cancellation and RAII cleanup (#187)

### LINKED ISSUE #187: Package install hangs after part download failures (no clean exit)
## Summary

When multi-part archive downloads in `xearthlayer packages install <region>` fail (any reason: network errors, retries exhausted, etc.), the program hangs indefinitely instead of returning a clean error and exiting. Users have to Ctrl-C to recover.

This is a defect independent of *why* downloads fail. Even when the failures themselves are environmental and unfixable, the install flow should always terminate cleanly.

## User-visible symptom

Quoting [#186](https://github.com/samsoir/xearthlayer/issues/186):

> When all download are finished or the error spawned, the program get stuck and nothing more happens. I need to ctrl-c to gett backto the shell.

## Root cause

Located in `xearthlayer/src/manager/download/strategy.rs:280-311` (`ParallelStrategy::execute()`):

1. **Blocking `handle.join()` with no timeout.** If any spawned worker thread fails to terminate (panic with unwound but un-released resources, deadlock on a shared `Mutex`, lingering tokio task holding a strong `Arc`), the main thread parks forever.

2. **`signal_done()` ordering.** The progress reporter background thread polls `is_done()` and only stops when `counters.signal_done()` is called — which happens **after** the join loop completes. While the main thread is parked on `join()`, the progress thread keeps polling, and from the user's perspective it looks like the whole program is hung.

## Proposed fix

Two complementary changes:

1. **Reorder signal_done()**: signal completion to the progress reporter *before* joining workers, not after. This way, the progress thread terminates regardless of join behavior.

2. **Add a join timeout**: wrap `handle.join()` with a deadline (e.g. 30 seconds after the last network activity, or based on `request_timeout_secs * max_retries`). On timeout, log the worker as orphaned, mark its part failed, and proceed. Workers that haven't returned by then are by definition stuck and won't recover.

Optional follow-up: instrument the join loop with structured logging so future hangs are diagnosable from logs alone.

## Validation

- Manual: simulate a hung worker (e.g. a test thread that never returns) and confirm the install command exits with an error within the join timeout window rather than parking indefinitely.
- Verify the progress reporter thread terminates even when worker threads hang (via `signal_done()` reorder).

## Related

- #186 — user report where this hang surfaced after EU package download errors. The download failures themselves are likely environmental (HTTP/2 over flaky link), but the hang is a separate defect.

samsoir:
## Summary

Reworks the multi-part download path so that a hard download failure aborts the install cleanly instead of letting the program hang. Three commits on this branch:

1. **2d45197** — `refactor(manager/download): consolidate on ParallelStrategy` — delete `SequentialStrategy` and the `DownloadStrategy` trait. SSOT consolidation; `concurrency=1` users now get retry + detailed progress.
2. **658e5be** — `refactor(manager/download): inline ParallelStrategy into MultiPartDownloader` — delete `strategy.rs`, lift the worker logic into `orchestrator.rs` as a private method. Removes load-bearing nostalgia.
3. **7da5c32** — `fix(packages): first-failure abort with cancellation and RAII cleanup` — the actual #187 fix.

## Behaviour spec (from issue body)

- Per-part: ≤3 attempts with exponential backoff (unchanged).
- On the first part to exhaust retries:
  - Set a shared `Arc<AtomicBool>` abort flag.
  - Other workers exit at their next abort check (after acquiring their semaphore permit, between attempts, during backoff via `interruptible_sleep`).
  - Temp directory wiped on every exit path via `InstallTempGuard` (RAII).
  - Non-zero exit with first-failure-specific error: `failed to download <filename>: <error> (N other parts also failed)`.

## Known limitation (called out in the commit message)

In-flight HTTP downloads are not interrupted mid-stream. Workers finish their current attempt before observing the abort. Plumbing cancellation into `reqwest` would be a separate, larger change. For dee-sea's reported scenario (many parts erroring quickly), this doesn't matter — failing workers hit the error path and check abort before the next retry.

## Tests
- `manager::install_guard`: 3 (drop wipes tree, drop silent on missing path, drop silent on partially-removed path)
- `manager::download::orchestrator`: 5 new (`interruptible_sleep` × 3, error-format singular vs plural)
- Full suite: 2398 passed, 9 ignored

## Test plan
- [x] `make pre-commit` passes locally
- [x] CI green
- [x] Manual integration validation by maintainer (induced URL failures): confirm clean Err + non-zero exit + temp dir cleaned
- [ ] Cross-check against dee-sea's `RUST_LOG=trace` log when it arrives on #186

## Discussion threads on the issue
- Diagnostic ask posted to dee-sea on #186 (verbose log + workaround attempt) so we can validate the fix matches the actual hang signature.
- Out-of-scope but noted in PR-side review: `services.rs:302-305` calls `bar.finish_and_clear()` after install. If indicatif's drawing logic ever blocks on a bar that never reached 100%, that would be a separate hang site. Not addressing here unless evidence appears.

Fixes #187


samsoir:
:shipit:

PR #192, opened 2026-05-09
chore(provider): replace manual JSON parsing in Google with serde (#163)

### LINKED ISSUE #163: chore: replace manual JSON parsing in Google provider with serde
## Summary

The Google Maps provider uses manual string splitting (`split("\"session\":")`) to extract tokens from API responses. This is brittle and will break if the JSON structure changes. Additionally, `GoogleMapsProvider` and `AsyncGoogleMapsProvider` share ~80% of their logic but are implemented as separate structs.

### Replace manual parsing
- Use `serde_json` (already in the dependency tree) for all API response parsing
- Define response structs with `#[derive(Deserialize)]`

### De-duplicate sync/async providers
- Extract shared logic (URL building, coordinate mapping, response parsing) into a shared inner struct or helper functions
- Sync and async variants delegate to shared core, differing only in HTTP client usage
- Apply the same pattern to other providers if they have similar duplication

## Files

- `xearthlayer/src/provider/google.rs`
- `xearthlayer/src/provider/async_google.rs` (if separate)

samsoir:
## Summary
- Swap the brittle `split("\"session\":")` chain in Google's `createSession` handler for a serde-derived `SessionResponse` struct. Non-string session values, missing fields, malformed JSON, and invalid UTF-8 now all fail loudly instead of being silently misinterpreted.
- Light de-dup of `GoogleMapsProvider` (sync) and `AsyncGoogleMapsProvider` (async): lift `SESSION_REQUEST_BODY`, `build_session_url`, `build_tile_url`, and `parse_session_response` to module-level free functions. Both wrappers delegate; the only remaining difference is the `.await` on `post_json`.
- No new dependencies — `serde_json` was already in the tree.

## Behavior change
The old split-based parser would silently misinterpret `{"session": 12345, "expiry":"X"}` by pulling `expiry` as the session token. The new parser rejects it via `Deserialize`'s type checking. New regression test (`parse_session_response_rejects_non_string_session`) pins this.

## Scope
Narrow to `google.rs` only. The "apply same pattern to other providers" line in #163 is dropped from this PR — each provider deserves its own scoped chore (e.g., `apple.rs` JWT handling).

## Test plan
- [x] 18/18 google provider unit tests pass (12 pre-existing + 6 new helper tests)
- [x] Full suite green via `make pre-commit` (2,404 tests)
- [x] Diff is +103/-69 in a single file; production code shrank, tests grew

Fixes #163

### DIFF
--- a/xearthlayer/src/provider/google.rs
+++ b/xearthlayer/src/provider/google.rs
@@ -26,7 +26,56 @@
 //! This differs from Bing's quadkey system but maps directly to our
 //! tile coordinates.
 
+use serde::Deserialize;
+
 use crate::provider::{AsyncHttpClient, AsyncProvider, HttpClient, Provider, ProviderError};
+
+/// Subset of the `createSession` response we care about.
+///
+/// Google returns additional fields (e.g. `expiry`, `tileWidth`); we ignore
+/// them and rely on serde to fail loudly if `session` is missing or non-string.
+#[derive(Deserialize)]
+struct SessionResponse {
+    session: String,
+}
+
+/// JSON body sent to `createSession` (satellite map type, en-US, US region).
+const SESSION_REQUEST_BODY: &str = r#"{
+  "mapType": "satellite",
+  "language": "en-US",
+  "region": "US"
+}"#;
+
+/// Build the URL used to mint a session token for the given API key.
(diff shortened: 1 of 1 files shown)


samsoir:
:shipit:

PR #196, opened 2026-05-11
chore(config): mask sensitive credentials in config list and diagnostics (#162)

### LINKED ISSUE #162: chore: mask sensitive config values in CLI output
## Summary

`xearthlayer config list` and `xearthlayer config get` display API keys and access tokens in plain text. These should be masked in terminal output.

## Affected Keys

- `provider.google_api_key` — Google Maps API key
- `provider.mapbox_access_token` — MapBox access token

## Proposed Implementation

- Add `is_sensitive()` method to `ConfigKey`
- Mask sensitive values in `config list` and `config get` output (e.g., `AIza...****`)
- Show full value only with an explicit `--reveal` flag if needed

## Notes

Low risk in practice — these are local CLI operations, not network-exposed. But good hygiene and a quick win (~30 min implementation).

samsoir:
## Summary

Two coordinated changes that give XEarthLayer a single source of truth for "what is a credential": `ConfigKey::is_sensitive`. Both `xearthlayer config list` and `xearthlayer diagnostics` now mask provider credentials (`provider.google_api_key`, `provider.mapbox_access_token`) consistently, so output from either is safe to paste into bug reports.

### 1. Mask in \`config list\` (\`4f9fd86\`)
- New \`ConfigKey::is_sensitive(&self) -> bool\` returning true only for the two provider credentials.
- New public \`SENSITIVE_VALUE_MASK\` constant (\`xxxxxxxx\`).
- \`xearthlayer config list\` substitutes the mask for any sensitive key with a configured value.
- \`config get\` and \`config set\` are intentionally left unmasked — they are authoritative user actions, and masking them would prevent users from inspecting or echoing back the value they just configured.
- Empty values still display as \`(not set)\` because absence isn't a leak.

### 2. Unify \`diagnostics\` redaction on the same predicate (\`de97135\`)
\`diagnostics/report.rs\` was previously using ad-hoc substring matching (\`api_key\`/\`token\`/\`secret\`) and a different mask (\`[REDACTED]\`), creating two competing SSOTs.

Replace the substring loop with section-aware INI parsing that looks each \`key = value\` line up via \`ConfigKey::from_str\` and masks via \`SENSITIVE_VALUE_MASK\` when \`is_sensitive()\` returns true.

**Behavioral changes from the swap:**
- Mask format \`[REDACTED]\` → \`xxxxxxxx\` (matches \`config list\`).
- A future \`[provider]\` credential added to \`ConfigKey\` is masked automatically, even if its name doesn't contain \`api_key\`/\`token\`/\`secret\`.
- A typo'd line like \`[cache] google_api_key = ...\` no longer triggers a false-positive redaction — it parses as an unknown ConfigKey and passes through. The exhaustive \`is_sensitive\` test in \`config::keys\` guarantees we won't miss real sensitive variants.

The redaction logic is extracted as a pure \`redact_sensitive_ini(&str) -> String\` helper so it's testable without filesystem I/O.

## Tests

9 new tests, all green via \`make pre-commit\` (2,413 total):
- \`config::keys\`: \`is_sensitive\` truthy/falsy + exhaustive guard so a future ConfigKey variant fails CI without an explicit sensitivity decision; mask constant pinned to \`xxxxxxxx\`.
- \`diagnostics::report\`: credential masking, empty-value pass-through, structural preservation (comments / headers / non-credentials), unknown keys, section-correctness.

## End-to-end manual verification

\`\`\`
\$ xearthlayer config set provider.google_api_key "AIzaTEST_SHOULD_NEVER_LEAK"
Set provider.google_api_key = AIzaTEST_SHOULD_NEVER_LEAK    # set echoes raw (authoritative)

\$ xearthlayer config get provider.google_api_key
AIzaTEST_SHOULD_NEVER_LEAK                                   # get shows raw (authoritative)

\$ xearthlayer config list | grep google_api_key
  google_api_key = xxxxxxxx                                  # list masks

\$ xearthlayer diagnostics | grep google_api_key
google_api_key = xxxxxxxx                                    # diagnostics masks too, same format
\`\`\`

Fixes #162

### DIFF
--- a/xearthlayer-cli/src/commands/config.rs
+++ b/xearthlayer-cli/src/commands/config.rs
@@ -6,7 +6,7 @@
 
 use clap::Subcommand;
 use xearthlayer::config::{
-    analyze_config, config_file_path, upgrade_config, ConfigFile, ConfigKey,
+    analyze_config, config_file_path, upgrade_config, ConfigFile, ConfigKey, SENSITIVE_VALUE_MASK,
 };
 
 use crate::error::CliError;
@@ -120,14 +120,22 @@ fn run_list() -> Result<(), CliError> {
             current_section = section;
         }
 
-        let value = key.get(&config);
+        let raw_value = key.get(&config);
         let key_name = key.key_name();
 
-        if value.is_empty() {
-            println!("  {} = (not set)", key_name);
+        // Mask credentials so the output is safe to paste into bug
+        // reports. `(not set)` shows through unmasked because absence
+        // isn't a leak and is informative for users debugging
+        // "why isn't my Google provider working." See issue #162.
+        let display_value = if raw_value.is_empty() {
+            "(not set)".to_string()
+        } else if key.is_sensitive() {
+            SENSITIVE_VALUE_MASK.to_string()
         } else {
-            println!("  {} = {}", key_name, value);
-        }
+            raw_value
+        };
+
+        println!("  {} = {}", key_name, display_value);
     }
 
     Ok(())
--- a/xearthlayer/src/config/keys.rs
+++ b/xearthlayer/src/config/keys.rs
@@ -11,6 +11,14 @@ use super::file::ConfigFile;
 use super::size::{format_size, parse_size};
 use crate::dds::DdsFormat;
 
+/// Display value used in place of any [`ConfigKey::is_sensitive`] value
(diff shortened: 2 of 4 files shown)


samsoir:
:shipit:

PR #202, opened 2026-06-12
Feat/macos port clean

dlicudi:
## Summary

Ports XEarthLayer to macOS. The branch is strictly portability: no feature
changes, no behavior changes on Linux. Tested on Apple Silicon (M5 Max,
macOS 26, macFUSE 5.2) against stable X-Plane 12.4.2 with multi-hour
flights and no session losses.

Developed with AI assistance (Claude); all changes flight-tested by me.

## Changes

- Mount the fuse3 backend on macFUSE, with unified unmount handling and
  mount verification (`system/unmount.rs`)
- Page-cache the virtual DDS files so X-Plane's mmap reads do not crash
- Force-unmount on shutdown so the mountpoint is not left stale
- Auto-recover stale FUSE mounts at startup (a run killed by crash or
  SIGKILL leaves a dead mount; detection must use opendir, not stat,
  because macOS serves cached attributes for dead macFUSE mountpoints)
- Platform-aware mount failure hints (macFUSE install/approval/umount
  on macOS instead of apt/fusermount)
- macOS location for the X-Plane install reference file
  (~/Library/Preferences/x-plane_install_12.txt)
- BSD-compatible `split` invocation in the publisher (probe and `-b`)
- `run` falls back to Custom Scenery auto-detection like the packages
  commands already did
- fuse3 pinned by rev to a fork carrying one reply-resilience patch,
  submitted upstream as Sherlock-Holo/fuse3#137 (rebased on upstream
  master, which already includes Sherlock-Holo/fuse3#136). Once
  Sherlock-Holo/fuse3#137 merges, one commit
  repoints to upstream and also retires the samsoir/fuse3 branch pin
  currently on main.
- Docs: platform support updated in README and CLAUDE.md

## Related Issues

- Related: https://github.com/Sherlock-Holo/fuse3/pull/137 (upstream fuse3
dependency, pending)
- Closes #201

## Test Plan

- [x] `make pre-commit` passes (fmt + clippy + tests)
- [x] Full test suite green on macOS (2,420 lib tests + integration)
- [x] Manual on macOS: `xearthlayer run --airport <ICAO>` mounts, streams
      tiles, and unmounts cleanly on Ctrl-C; 2h and 4h+ real X-Plane
      flights without crashes
- [x] Stale mount recovery verified end-to-end: mount, kill -9, restart,
      observe "Stale FUSE mount detected" warning and healthy remount
- [ ] Regression check on Linux (no Linux hardware on my side; CI covers
      build and tests)

## Checklist

[...]

Issue #201, opened 2026-06-03
[Feature]: macOS (macFUSE) support - working spike branch below


dlicudi:
### Proposed Solution

A working macFUSE port exists on a spike branch in my fork
(dlicudi/xearthlayer), using the fuse3 async backend, and flies in X-Plane 12.
Most recent test on 12.4.3 b4: a ~113 minute flight from LEMG. xearthlayer and
the mount stayed healthy the whole time; X-Plane itself crashed at ~113 min (an
X-Plane side stack overflow, see Known issue).

Links:
- Tested snapshot (tag): https://github.com/dlicudi/xearthlayer/tree/macos-spike-2026-06-03
- Ongoing branch: https://github.com/dlicudi/xearthlayer/tree/feat/macos-port-spike
- fuse3 fork: https://github.com/dlicudi/fuse3/tree/fix/macos-reply-error-resilience

Environment:
- macOS 26.4.1 (Apple Silicon, Metal)
- X-Plane 12.4.3-b4 (build 124307)
- macFUSE
- Provider: Bing, consolidated ortho union mount

What works:
- fuse3 mounts on macFUSE, X-Plane reads on-demand DDS tiles
- Three-tier cache, prefetch, Web API telemetry, prewarm
- Mount stays up across a long session, including bursts of macFUSE EINVAL
  replies that previously tore it down

Requires a small fuse3 fork: samsoir/fuse3@configurable-background plus one
patch so macFUSE EINVAL replies do not tear the mount down.

Known issue: X-Plane crashes during long flights (~40 min on beta 3, ~113 min
on beta 4). Triaged from the X-Plane minidump: a stack overflow from infinite
recursion in an X-Plane background worker thread (EXC_BAD_ACCESS /
KERN_PROTECTION_FAILURE, ~13,400 recursive frames). No XEL or macFUSE frames
appear in the stack, and xearthlayer plus the mount stay healthy across the
crash, so this is an X-Plane side bug, possibly triggered by the lazily served
textures, not a fault in XEL.

dlicudi:
Experimenting with XEL_SERVE_PLACEHOLDER_ON_MISS: on a cache miss taking >50ms, serve a magenta placeholder immediately and keep rebuilding the tile async (not cancelled).

Goal: keep X-Plane stable by avoiding tile-serving delays at all costs.

[...]

PR #211, opened 2026-07-30
chore(ci): add macOS build, test and release support

samsoir:
## Summary

Makes macOS a **Tier 1 platform** in the CI/CD pipeline, ahead of merging #202 (the macOS port), and brings every version-carrying file on `develop/0.5.0` into `0.5.x` agreement.

> **Expected red check.** `Verify (macOS)` fails on this branch because `develop/0.5.0` contains no `cfg(target_os = "macos")` code — that all arrives with #202. This is deliberate. Because `pull_request` workflows run from the merge ref, **#202's own run compiles `develop/0.5.0` + #202 together and should go green** — which is precisely the automated macOS evidence this branch exists to provide. The red window closes when #202 merges.

## Changes

- **CI** — new blocking `Verify (macOS)` job on pinned `macos-15` in `ci.yml` and `release-test.yml`. The existing `Verify` job is deliberately untouched: `main`'s branch protection requires that *exact* status context, and GitHub renames matrix jobs (`Verify (ubuntu-latest)`), so a matrix would leave merges to `main` waiting forever on a check that never reports.
- **Release** — `verify-macos` + `package-macos` jobs. `publish` is gated on `package-macos` in **both** its `needs:` list and its `if:` expression, because that gate uses `!cancelled()` plus per-job result checks so it still runs when the stable-only jobs are skipped on a pre-release tag — `needs:` alone would not block publication. Ships `xearthlayer-<tag>-arm64-macos.tar.gz` on both channels. Build and package are fused into one job: `actions/upload-artifact` does not preserve the executable bit, so the tarball is created before upload rather than `chmod`'d after download.
- **`make verify-macos`** — guards on macOS and macFUSE, runs `make verify`, then the live `macfuse_smoke` tests that CI structurally cannot execute (macFUSE is a kernel extension; hosted runners cannot load one). This is the human half of the Tier 1 gate.
- **`make bump-version` repaired** — GNU-only `sed -i` replaced with a portable rewrite (it was simply broken under BSD sed on macOS); coverage widened from 2 files to 5; two version forms, since RPM uses `-` to delimit `Version` from `Release` and Arch `pkgver` forbids hyphens:

  | Target | Form |
  |---|---|
  | `Cargo.toml`, `Cargo.lock`, `version.json` | full semver (`0.5.0-dev.0`) |
  | `pkg/rpm/xearthlayer.spec`, `pkg/arch/PKGBUILD` | hyphen-stripped base (`0.5.0`) |

  It also validates the Fedora tag (`fcNN`) read back from `version.json` **before writing any file**, so a malformed value aborts cleanly instead of leaving a half-bumped tree.
- **Version sweep** — `version.json` `0.4.6` → `0.5.0-dev.0`, `pkg/rpm/xearthlayer.spec` `0.2.5` → `0.5.0`, `pkg/arch/PKGBUILD` `0.2.0` → `0.5.0`. Those last two had drifted invisibly: CI rewrites the RPM `Version:` at build time and generates the AUR `PKGBUILD` from a heredoc, so their staleness never surfaced — while `make pkg-rpm`, which does *not* rewrite, was producing an RPM labelled `0.2.5`. The root cause was `bump-version` never knowing those files existed, which is why the tool fix and the sweep are both here.
- **Docs** — new `docs/dev/cicd.md` (pipeline and merge-strategy reference: branch model, release job graph, platform tiers, release channels, status checks, version propagation). Runbook, `CONTRIBUTING.md`, `CLAUDE.md` and `CHANGELOG.md` updated. The runbook's release procedures now point at `make bump-version` and stage all five files — three `git add` lists were silently invalidated when the tool's coverage widened.

## Related Issues

- Related: #201 (macOS support)
- Related: #202 (macOS port — this branch is its CI/CD prerequisite and provides the automated macOS evidence for approving it)

## Test Plan

- [x] `make pre-commit` passes (fmt + clippy + tests) — ~2398 lib tests, 0 failures, at the branch tip
- [x] `bump-version` idempotence: re-running at the current version yields an **empty** diff. Because `release_date` is deliberately left manual, this doubles as a drift detector
- [x] Fedora-tag guard verified *firing*: missing `.assets.rpm`, malformed JSON, filename with no `fcNN`, and empty file all abort non-zero with `version.json` left unmodified
- [x] `make verify-macos` platform guard exits non-zero on Linux in milliseconds without invoking cargo
- [x] Workflow YAML validated with `actionlint`; the `publish` DAG and `if:` gate asserted programmatically against the parsed YAML
- [x] macOS tarball filename cross-checked between `release.yml`, `bump-version`'s `jq` filter, and `version.json` — all three agree, so no 404 download link
- [ ] **Requires Apple Silicon** — `make bump-version` under BSD sed, and `make verify-macos`'s full path (needs macFUSE installed plus `tests/macfuse_smoke.rs`, which arrives with #202)
- [ ] **First `v0.5.0-dev.N` tag** — confirm the macOS tarball appears as a release asset and that `publish` waited on `package-macos`

## Checklist

- [x] Code follows the project's SOLID principles and patterns — no Rust source changed
- [x] Documentation updated if applicable (CLAUDE.md, docs/, config reference)

## Follow-ups (not in this PR)

1. **Branch protection on `develop/0.5.0`** requiring `Verify` + `Verify (macOS)` — must wait until #202 merges, since protecting the branch while that job is red would block every merge including #202 itself.
2. **`docs/macos.md` Gatekeeper note** (`xattr -dr com.apple.quarantine`) — that file arrives with #202.
3. **Website `macos_tarball` support** — `sync-version.yml` reads only `deb`/`rpm`/`tarball`, so the macOS asset ships with no download link on the site. Verified non-breaking (unknown keys are ignored, as `aur` already is), but needs tracking.
4. **Website `aur` shortcode** renders a broken URL — pre-existing; `getting-started.md` references an asset key the sync never writes.
5. **Homebrew tap** — would sidestep Gatekeeper entirely by building from source.
6. **`make pkg-tarball` vs CI tarball naming** disagree on the `v` prefix — pre-existing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,3 +37,26 @@ jobs:
       - name: Run make verify
         run: make verify
 
+  # macOS is a Tier 1 platform. Pinned to macos-15 (Apple Silicon) rather than
+  # macos-latest: that label has moved across both OS versions and CPU
+  # architectures, and a silent bump would surface as an unexplained failure on
+  # an unrelated PR.
+  #
+  # No dependency-install step is needed. The pinned fuse3 fork is pure Rust
+  # (no build.rs / pkg-config / cc) and xearthlayer/build.rs already emits
+  # -lc++ for macOS. macFUSE is a runtime dependency only, and cannot be loaded
+  # on hosted runners in any case (it is a kext).
+  verify-macos:
+    name: Verify (macOS)
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt, clippy
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Run make verify
+        run: make verify
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -32,3 +32,28 @@ jobs:
 
       - name: Run make verify
         run: make verify
+
+  # macOS is a Tier 1 platform. Pinned to macos-15 (Apple Silicon) rather than
+  # macos-latest: that label has moved across both OS versions and CPU
+  # architectures, and a silent bump would surface as an unexplained failure on
+  # an unrelated PR.
+  #
+  # No dependency-install step is needed. The pinned fuse3 fork is pure Rust
+  # (no build.rs / pkg-config / cc) and xearthlayer/build.rs already emits
+  # -lc++ for macOS. macFUSE is a runtime dependency only, and cannot be loaded
+  # on hosted runners in any case (it is a kext).
+  verify-macos:
+    name: Verify (macOS)
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt, clippy
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Run make verify
+        run: make verify
+
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,23 @@ jobs:
       - name: Run make verify
         run: make verify
 
+  # macOS release gate. Mirrors the Linux `verify` job; see ci.yml for why the
+  # runner is pinned and why no dependency install is required.
+  verify-macos:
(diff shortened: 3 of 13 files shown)


samsoir:
macOS builds are expected to fail on this PR, will be fixed in PR #202

PR #214, opened 2026-08-03
fix(dds): emit complete mipmap chains in generated DDS

### LINKED ISSUE #212: Moiré/striped pattern on distant ZL16 tiles near LSGG – disappears when zooming in
### Bug Description

I've discovered a strange artifact: Here near Geneva (LSGG), when I simply fly from the airport into the sky, I see a pattern in the mountains in the rendering; if I zoom in, the pattern disappears. 

I don't see these artifacts when I use my own ortho tiles. OK, my own tiles have ZL 17, and I suspect XEarthLayer uses ZL 16 for the distance... but is that the reason a pattern is generated in the display? Or is it due to some graphics setting that generates this pattern at ZL 16 but not at ZL 17? I’ve already tried all the settings in X-Plane....

I also tried my own ortho XEarthLayer package for the Alps—still no difference. Only when using complete ortho tiles do I not see these stripes.

I’m a bit at a loss as to what other settings or switches I could try.

Attached are a few screenshots showing the striped pattern, and when you zoom in, you can see it without the stripes. Just zoom with the mouse wheel....

### Steps to Reproduce

Steps to reproduce:

Start XEarthLayer
Start X-Plane at airport LSGG (Geneva)
Take off / increase altitude to get a view into the Alps
Observe the striped pattern on the distant mountains
Zoom in with the mouse wheel → the pattern disappears

Image

Image

Image

### Expected Behavior

no arttifact at all

### Actual Behavior

see attachments

### System Diagnostics

```text
XEarthLayer Diagnostics
=======================

XEarthLayer Version: 0.4.6

## Operating System
Kernel: 7.1.5-1-liquorix-amd64
OS: Debian GNU/Linux 13 (trixie)
Desktop: KDE
Display Server: wayland

## Hardware
CPU: AMD Ryzen 7 9800X3D 8-Core Processor (16 cores)
Memory: 92.0 GB

## GPU
GPU: NVIDIA Corporation AD102 [GeForce RTX 4090] (rev a1)
GPU: Advanced Micro Devices, Inc. [AMD/ATI] Granite Ridge [Radeon Graphics] (rev cb)
VRAM (NVIDIA): 24564 MB (24.0 GB)

## GPU Compute Adapters
  [0] NVIDIA GeForce RTX 4090 (Discrete, vulkan)
  [1] AMD Radeon Graphics (RADV RAPHAEL_MENDOCINO) (Integrated, vulkan)
  [2] llvmpipe (LLVM 19.1.7, 256 bits) (Software fallback, vulkan)
  [3] NVIDIA GeForce RTX 4090/PCIe/SSE2 (Other, gl)

## Disk Space
Cache directory: /mnt/xplane_data/xearthlayer/cache (378G used, 4531.4 GB available)
Packages location: /home/maechtel/.xearthlayer/packages (681.7 GB available)
Root filesystem: 165G used of 504G (35%)
Config directory: exists

## Network
Default interface: enp11s0

## XEarthLayer Configuration
Config file: /home/maechtel/.xearthlayer/config.ini


[general]
; Check for new versions on startup (default: true)
; When enabled, performs a single HTTP request once per day (no telemetry)
update_check = true

[provider]
; Imagery provider:
;   apple  - Apple Maps (free, tokens auto-acquired via DuckDuckGo)
;   arcgis - ArcGIS World Imagery (free, global coverage)
;   bing   - Bing Maps (free, no key required)
;   go2    - Google Maps via public tile servers (free, no key required, same as Ortho4XP)
;   google - Google Maps official API (paid, requires API key)
;   mapbox - MapBox satellite (free tier available, requires access token)
;   usgs   - USGS orthoimagery (free, US coverage only)
type = bing
; Google Maps API key (only required when type = google)
; Get one at: https://console.cloud.google.com (enable Map Tiles API)
google_api_key =
; MapBox access token (only required when type = mapbox)
; Get one at: https://www.mapbox.com/
mapbox_access_token =

[cache]
; Base directory for disk cache storage. Chunks are stored in <directory>/chunks/
; If empty, defaults to ~/.cache/xearthlayer (Linux) or platform cache directory
; Example: directory = /mnt/fast-ssd/xearthlayer
directory = /mnt/xplane_data/xearthlayer/cache
; Memory cache size (default: 2GB) - uses RAM for fastest access
; Supports: KB, MB, GB suffixes (e.g., 500MB, 2GB, 4GB)
memory_size = 4 GB
; Disk cache size (default: 20GB) - persistent storage for tiles
; Supports: KB, MB, GB suffixes (e.g., 10GB, 20GB, 50GB)
disk_size = 1500 GB
; Fraction of disk cache allocated to DDS tiles (default: 0.6)
; Remainder goes to raw imagery chunk cache. Range: 0.0 - 1.0
dds_disk_ratio = 0.6
; Disk I/O concurrency profile based on storage type (default: auto)
;   auto - Auto-detect storage type (recommended)
;   hdd  - Spinning disk (conservative: 1-4 concurrent ops)
;   ssd  - SATA/AHCI SSD (moderate: ~32-64 concurrent ops)
;   nvme - NVMe SSD (aggressive: ~128-256 concurrent ops)
disk_io_profile = auto

[texture]
; DDS compression format: bc1 (smaller, opaque) or bc3 (larger, with alpha)
; bc1 recommended for satellite imagery
format = bc1
; Texture compressor backend: software, ispc (SIMD), or gpu (wgpu)
; ispc is recommended for best CPU performance
compressor = gpu
; GPU device selector for gpu compressor: integrated, discrete, or adapter name substring
gpu_device = integrated

[generation]
; Number of threads for parallel tile generation (default: num_cpus / 2)
; WARNING: Do not set this higher than your CPU core count
threads = 8
; Timeout in seconds for generating a single tile (default: 10)
; If exceeded, returns a magenta placeholder texture
timeout = 10

[executor]
; Job executor daemon settings for tile generation.
; These control resource pools, concurrency, and retry behavior.

; Resource pool capacities (concurrent operations by type)
; Network: HTTP connections for chunk downloads (default: 128, clamped to 64-256)
network_concurrent = 128
; CPU: Assemble + encode operations (default: num_cpus / 2)
cpu_concurrent = 16
; Disk I/O: Cache read/write operations (default: 64 for SSD)
disk_io_concurrent = 128
; Maximum concurrent DDS tile jobs (default: num_cpus / 2)
max_concurrent_jobs = 8

; Download behavior
; HTTP request timeout in seconds for individual chunk downloads (default: 10)
request_timeout_secs = 10
; Maximum retry attempts per failed chunk download (default: 3)
max_retries = 3
; Base delay in milliseconds for exponential backoff between retries (default: 100)
; Actual delay = base_delay * 2^attempt (e.g., 100ms, 200ms, 400ms, 800ms)
retry_base_delay_ms = 100

[xplane]
; X-Plane Custom Scenery directory for mounting scenery packs
; If empty, auto-detects from ~/.x-plane/x-plane_install_12.txt
; This is also where packages are installed by 'xearthlayer packages install'
scenery_dir = ~/X-Plane-12/Custom Scenery

[packages]
; URL to the XEarthLayer package library index
; This is where available packages are discovered for 'xearthlayer packages check/install'
library_url = https://xearthlayer.app/packages/xearthlayer_package_library.txt
; Local directory for installed packages (default: ~/.xearthlayer/packages)
install_location = ~/.xearthlayer/packages
; X-Plane Custom Scenery directory for overlay symlinks
; If empty, uses xplane.scenery_dir or auto-detects
custom_scenery_path = ~/X-Plane-12/Custom Scenery
; Automatically install overlay packages when installing ortho for same region
auto_install_overlays = false
; Suppress XEL overlays at runtime by removing the consolidated yzXEL_overlay/
; symlink folder from Custom Scenery on `xearthlayer run` startup. Overlay
; packages are still downloaded and stored locally — only the symlink that
; X-Plane reads is suppressed. Use this when running third-party overlay
; scenery (e.g., SimHeaven) that conflicts with XEL overlays. (default: false)
disable_overlays = false
; Temporary directory for package downloads (default: ~/.xearthlayer/tmp)
; Large packages are downloaded here before extraction
temp_dir =
; Number of concurrent part downloads (1-10, default: 5)
concurrent_downloads = 10

[logging]
; Log file path (default: ~/.xearthlayer/xearthlayer.log)
file = ~/.xearthlayer/xearthlayer.log

[prefetch]
; Enable predictive tile prefetching based on X-Plane telemetry (default: true)
; Requires X-Plane to send ForeFlight data: Settings > Network > Send to ForeFlight
enabled = true
; Adaptive prefetch mode (default: auto)
;   auto         - Select mode based on throughput calibration (recommended)
;   aggressive   - Position-based triggers (fast connections)
;   opportunistic - Circuit breaker triggers (moderate connections)
;   disabled     - Disable prefetch entirely
mode = auto
; Web API port for X-Plane SimState polling (default: 8086, range: 1024-65535)
web_api_port = 8086

; Interval between prefetch cycles in milliseconds (default: 2000)
; Higher values reduce prefetch aggressiveness
cycle_interval_ms = 4000

; Adaptive prefetch calibration
; Minimum throughput for aggressive mode (tiles/sec, default: 30)
; Systems exceeding this use position-based prefetch triggers
calibration_aggressive_threshold = 30
; Minimum throughput for opportunistic mode (tiles/sec, default: 10)
; Systems between this and aggressive use circuit breaker triggers
; Below this, prefetch is disabled
calibration_opportunistic_threshold = 10
; How long to measure throughput during initial calibration (seconds, default: 60)
calibration_sample_duration = 60

; Transition ramp settings (takeoff phase management)
; Altitude climb (feet) above takeoff MSL to release transition hold (default: 1000, range: 200-5000)
takeoff_climb_ft = 1000
; Maximum seconds before timeout release if climb not reached (default: 90, range: 30-300)
takeoff_timeout_secs = 90
; Sustained seconds at GS < 40kt before landing detection (default: 15, range: 5-60)
landing_hysteresis_secs = 15
; Duration (secs) of linear ramp from start fraction to full rate (default: 30, range: 10-120)
ramp_duration_secs = 30
; Starting prefetch fraction when ramp begins (default: 0.25, range: 0.1-0.5)
ramp_start_fraction = 0.25

; Boundary-driven prefetch settings
; Buffer tiles for retention beyond visible window (default: 1, range: 0-3)
window_buffer = 1
; InProgress staleness timeout in seconds (default: 120, range: 30-600)
stale_region_timeout = 120

; Prefetch box settings (used by both ground and cruise phases)
; Total prefetch box extent per axis in degrees (default: 7.0, range: 3.0-15.0)
; On the ground: used directly with symmetric bias, producing a square box centered on the aircraft.
; In cruise: the maximum of a speed-proportional ramp from box_min_extent, with heading bias.
box_extent = 6.5
; Maximum forward bias fraction (default: 0.8, range: 0.5-0.9)
; Controls how much the prefetch box shifts forward in the direction of travel
; 0.5 = symmetric, 0.8 = 80% ahead / 20% behind on the primary axis
box_max_bias = 0.8

[prewarm]
; Settings for cold-start cache pre-warming.
; Use with --airport ICAO to pre-load tiles around an airport before flight.

; Grid rows (latitude extent) in DSF tiles around the airport to prewarm (default: 3)
grid_rows = 3
; Grid columns (longitude extent) in DSF tiles around the airport to prewarm (default: 4)
grid_cols = 4

[patches]
; Settings for custom Ortho4XP tile patches (airport addon mesh/elevation support).
; Patches provide custom elevation/mesh data while XEL generates textures dynamically.

; Enable/disable patches functionality (default: true)
; When enabled, XEL will mount patch tiles from the patches directory.
enabled = true
; Directory containing patch tiles (default: ~/.xearthlayer/patches)
; Each subdirectory should be a complete Ortho4XP tile with:
;   - Earth nav data/*.dsf (custom mesh/elevation)
;   - terrain/*.ter (terrain definition files)
;   - textures/ (optional - XEL generates these on-demand)
; Priority is determined by alphabetical folder naming (A < B < Z)
directory = ~/.xearthlayer/patches

[fuse]
; FUSE kernel settings for concurrent background request limits.
; Higher values allow more concurrent X-Plane scenery reads, preventing
; freezes at DSF boundaries. Only modify if you understand FUSE internals.

; Maximum pending background FUSE requests before the kernel queues (default: 256, range: 1-1024)
; The Linux kernel default of 12 severely limits X-Plane's concurrent scenery reads.
max_background = 128
; Congestion threshold for background FUSE requests (default: 192, range: 1-1024)
; Kernel starts throttling when pending requests exceed this. Convention: 75% of max_background.
congestion_threshold = 96
```

### Relevant Logs

```text

```

### Screenshots

_No response_

### Severity

Crash / Data Loss

### Checklist

mmaechtel:
Fixes #212.

## Cause

Generated DDS tiles declared and carried a **5-level mipmap chain** where a 4096×4096 texture supports **13**. X-Plane loads only the declared levels and clamps sampling at the last one. Past the distance at which the hardware would request a coarser level, the texture is undersampled rather than filtered. At grazing angles the texel footprint is compressed perpendicular to the fall line, so the aliasing beats against the terrain relief and reads as regular banding along the contours — invisible up close, independent of provider and zoom level.

```
xearthlayer   4096×4096 BC1   mipMapCount = 5    11,174,016 bytes
Ortho4XP      4096×4096 BC1   mipMapCount = 13   11,184,952 bytes
```

The eight missing levels are 10,936 bytes per tile — **+0.1 %**.

This also explains why the issue looked like a ZL16/ZL18 overlap problem: the artifact stands out most where texture scale changes, so the boundaries were simply where it was noticed first.

## Change

The level count is now derived from the texture dimensions at encode time via `MipmapGenerator::full_chain_count()` instead of a constant. `TextureConfig` and `DdsTextureEncoder` carry `Option<usize>`, where `None` means the full chain. An explicit count remains available for tests and is **clamped to the real chain length**, so the header can never over-declare what the payload holds. No `config.ini` setting is exposed — a truncated chain has no legitimate use case.

Four independent hardcoded `5`s were removed (config defaults, the `DdsTextureEncoder` constructor, the CLI `run` command, the FUSE placeholder); any one of them alone would have kept the old behaviour alive.

Two things had to move in lockstep:

- **`EXPECTED_DDS_SIZE`** gates every FUSE read on exact length equality. Left at the 5-level size it would have rejected every correctly generated tile and served magenta instead. It is now derived from the chain.
- **Sub-block mipmap levels.** The tail of a full chain reaches 2×2 and 1×1, below one BCn block. The GPU kernel asserts both dimensions are multiples of 4, and the worker's panic handler drains the queue and breaks its loop — so a failure there would disable GPU encoding for the rest of the process lifetime, not just fail one tile. The ISPC kernel reads a full 4×4 footprint and would run off the end of a 16-byte surface. Levels are now padded up to block size once, centrally, before compression.

## Tests

Regression tests walk the payload level by level, asserting the header count matches what was written, that the chain terminates at 1×1, and that no bytes remain — across square, non-square and non-power-of-two inputs, plus clamping and BC3.

Two GPU tests (`--ignored`, require hardware) cover what unit tests cannot reach: a 4096² tile yielding 13 levels at 11,184,952 bytes, and a chain through the 2×2/1×1 tail **followed by a second successful request**, which is what actually proves the worker survived rather than having silently exited. The existing `gpu_pipeline_stress` test requests a single level and never touched this.

## Verification

Live in X-Plane 12 at LSGS: 851/851 freshly generated tiles at `mipMapCount = 13` and exactly 11,184,952 bytes. No timeouts, no placeholders, no GPU errors; X-Plane logged no texture or DDS complaints. A/B against the pre-fix build at the same position, time of day and viewing angle: banding gone, no visible change up close.

## Notes

- **Performance is not fully tuned.** Measured on integrated graphics, 4096² BC1, median of 7 runs after warmup: 5 levels 54.54 ms, 13 levels 54.73 ms — the eight added levels cost 0.19 ms, within run-to-run spread. The CPU backends have not been profiled.
- **Cached tiles are stale.** Existing DDS blobs carry the old 5-level layout and are structurally valid, so nothing detects them automatically. `xearthlayer cache clear` is required after updating. No cache-format version key is included in this PR.


mmaechtel:
@samsoir this one is ready for review whenever you have time — no open TODOs on my side.

Branch is clean and mergeable, `make pre-commit` passes, and the fix is verified in X-Plane 12 as described above. The two GPU tests are `--ignored` and need real hardware, so they will not run in CI.

One thing worth flagging for the release notes: existing caches hold the old 5-level textures and will keep showing the artifact, so users need `xearthlayer cache clear` after updating.

samsoir:
Hi @mmaechtel I will take a look tonight. Can you retarget this fix against develop/0.4.7 please as this will become part of the next main release. Thank you kindly for your investigation here and work to fix.

mmaechtel:
Done! The branch is rebased onto `develop/0.4.7` and the PR base is retargeted accordingly. The rebase was clean — no conflicts, since 0.4.7 only adds the version bump and release docs on top of main, which don't overlap with this fix. Diff is unchanged: same two commits, ready for review.

Issue #209, opened 2026-07-27
[Bug]: OOM kill after 12h long-haul — 64 GB unbounded anonymous memory growth


samsoir:
## Bug Description

After a 12-hour long-haul flight test on v0.4.6, the `xearthlayer` process was terminated by the kernel OOM killer. There was no application-level error — the process was operating normally and the kernel killed it. Nothing in `xearthlayer.log` indicates a problem; the last log line is the instant of the kill.

The process had grown to **64.0 GB of anonymous memory** over 11h59m (~5.3 GB/hour, monotonic), against a configured memory cache of 4 GB. Roughly **60 GB is unaccounted for by design**.

This is a hard blocker for long-haul operation. Any flight beyond ~8-10 hours on a 64-96 GB machine will hit this.

### Evidence: kernel OOM report

`dmesg` is restricted on this host; `journalctl -k` has the record:

```
Jul 27 09:54:56 kernel: kswapd0 invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0
Jul 27 09:54:56 kernel: [  16295]  1000 16295 18552193  2435583  2431383     4200         0 137105408 14337841           200 xearthlayer
Jul 27 09:54:56 kernel: Out of memory: Killed process 16295 (xearthlayer)
                        total-vm:74208772kB, anon-rss:9725532kB, file-rss:16800kB
Jul 27 09:54:56 kernel: Free swap  = 176kB
Jul 27 09:54:56 kernel: Total swap = 96393212kB
Jul 27 09:55:01 kernel: oom_reaper: reaped process 16295 (xearthlayer)
```

Memory at time of kill:

| Process | Anonymous total | Resident | Swapped | VM |
|---|---|---|---|---|
| **xearthlayer** | **64.0 GB** | 9.3 GB | 54.7 GB | 70.8 GB |
| X-Plane 12 (`Main Thread`, pid 62756) | 33.1 GB | 6.0 GB | 27.1 GB | 52.8 GB |
| everything else combined | < 6 GB | | | |

Combined ~97 GB of anonymous memory on a 91 GiB machine. The OOM was arithmetically unavoidable.

### Run statistics (parsed from the 3.4 GB session log)

- Duration: `2026-07-27 04:55:09Z` → `16:54:55Z` = **11h 59m**
- Jobs: **120,999 submitted == 120,999 started == 120,999 completed** — no job backlog
- Chunk downloads: 25,647,104 successful
[...]

PR #215, opened 2026-08-07
feat(metrics): periodic memory telemetry and log verbosity cleanup

samsoir:
## Summary

Adds the instrument needed to diagnose the OOM kill in #209, and cuts the log verbosity that would otherwise bury it.

A 12-hour flight ended with the kernel OOM-killing the process at 64 GB of anonymous memory against a configured 4 GB memory cache. There was no application error — the last log line is the instant of the kill. Three candidate causes remain live and undiscriminated. This branch makes the next long flight produce a trace that tells them apart, without anyone needing to set environment variables or fly a controlled route.

The core idea: a once-per-minute `Memory sample` INFO line carrying process memory alongside the cache, GC and load state needed to interpret it. `disk_writes_active` and `mem_cache_writes_active` correlated against `rss_mb` and `swap_mb` are what discriminate the candidates.

```
2026-08-06 06:20:33Z  INFO Memory sample uptime_s=60 rss_mb=873 vm_mb=1459 swap_mb=0
  threads=71 tiles_done=0 encodes_active=0 chunks_ok=0 chunks_failed=0 mem_cache_mb=0
  dds_disk_mb=158268 chunk_disk_mb=55301 gc_evicted_mb=0 chunk_index_entries=3803083
  disk_writes_active=0 mem_cache_writes_active=0
```

## Changes

- **`MemoryProbe` trait + `ProcessMemoryProbe`** (`metrics/memory_probe.rs`) backed by the `memory-stats` crate — the only new dependency. Injected into `MetricsDaemon`, so the daemon stays testable without reading real process memory.
- **16-field `Memory sample` line** emitted on a fixed 60-second tick. Deliberately not configurable: traces get pooled across users and machines, and a fixed cadence removes one variable.
- **Two new metrics.** `chunk_index_entries` exposes chunk LRU index size (~12 M entries during the OOM flight, one of the candidate causes). `mem_cache_writes_active` closes a diagnostic hole — see below.
- **Allocator-environment line at startup**, recording `MALLOC_*` overrides when set, so a trace taken with them active is never silently compared against one without.
- **Six high-volume `info!` sites demoted to `debug!`.** `Chunk download success` alone was 3,803,083 lines — 98.6% of a 501 MB log from a 1h29m flight. A 12-hour flight drops from roughly 4 GB to roughly 50 MB, small enough to attach to an issue.
- **`docs/dev/memory-telemetry.md`** — field reference, a decision table for interpreting a trace, and a confounder section. Cross-referenced with `memory-profiling.md`, which cannot capture an OOM kill because heaptrack loses all data on `kill -9`.

### Two bugs found and fixed along the way

- **Upstream race in `memory-stats` 1.2.0.** Its Linux path guards one-time init with a CAS; the thread that *loses* skips init entirely and reads `SMAPS_EXIST` before the winner stores it, falling through to the STATM path with `PAGE_SIZE` still `0`. Result: a silent `MemoryStats { 0, 0 }`. In a memory trace that reads as "the process used no memory" — precisely the investigation this exists to serve. Closed with a `std::sync::Once` warm-up; verified over 20 consecutive runs after failing 2 in 6 before.
- **False doc comment** on `disk_writes_active` claiming DDS writes don't emit write events. They do (`build_and_cache_dds.rs:294`/`:307`). That field is the primary discriminator, and the comment would have led a reader to dismiss the leading candidate.

### Scope note

The final review found the originally-designed 14 fields could point at the *wrong* cause, so two were added:

- **`swap_mb`** — 54.7 GB of the 64 GB in #209 was swapped out, so an RSS-only trace would have shown a flat ~9 GB and read as healthy. `/proc/self/status` was already being opened for `Threads:` and carries `VmSwap:` a few lines away.
- **`mem_cache_writes_active`** — the memory-cache `tokio::spawn` in `build_and_cache_dds.rs` emitted no start/complete pair, so a backlog concentrated there presented as "rss climbing, gauges flat" and would have been read as allocator retention rather than a write backlog.

## Related Issues

Related: #209

## Test Plan

- [x] `make pre-commit` passes (fmt + clippy + tests) — 2420 lib tests, 78 CLI, 63 doctests, 0 failures
- [x] Field-assertion test verified by **mutation**: swapping a field's source (`mem_cache_mb` → `dds_disk_cache_size_bytes`) makes it fail; restoring makes it pass. Every numeric field is seeded with a pairwise-distinct value, asserted distinct at runtime, so a swap between any two cannot pass silently
- [x] MB truncation asserted with non-whole-MB byte counts, so raw bytes leaking under an `_mb` name fails
- [x] `memory-stats` race fix verified over 20 consecutive runs (was 2 failures in 6)
- [x] Live smoke check on real hardware with a FUSE mount: allocator line plus `Memory sample` lines 60 s apart, all fields populated
- [x] Log demotion verified at source level — none of the six messages remain at `info!`; `Chunk download error`/`failed` deliberately left at `warn!`

Reviewers can verify the live behaviour with:

```bash
cargo build --release
MALLOC_ARENA_MAX=2 ./target/release/xearthlayer run
# after ~70s:
grep -E "Allocator environment|Memory sample" ~/.xearthlayer/xearthlayer.log
```

## Checklist

- [x] Documentation updated if applicable (CLAUDE.md, docs/, config reference)

## Known follow-up (not addressed here)

`metrics/daemon.rs` adds every `DiskWriteCompleted { bytes }` into `state.chunk_disk_bytes_written`, including ~11 MB DDS writes, so DDS bytes inflate a chunk-tier counter that feeds the TUI throughput display. Pre-existing, deliberately left untouched — changing it would alter TUI numbers and belongs in its own issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -446,5 +446,6 @@ CI will fail if pre-commit checks were not run.
 - Zoom level overlap management: `docs/dev/zoom-level-overlap-design.md` (dedupe, gap analysis)
 - **Consolidated FUSE mounting**: `docs/dev/consolidated-mounting-design.md` (single ortho mount, patches + packages)
 - **GeoIndex design**: `docs/dev/geo-index-design.md` (geospatial reference database, patch region ownership)
+- **Memory telemetry**: `docs/dev/memory-telemetry.md` (periodic memory sampling, trace interpretation, confounders)
 - memorize review allow(dead_code) macros at major checkpoints. Refactor aggresively to remove them when appropriate.
 - memorize ensure to update the projects documentation to reflect the current state of the project before committing changes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,7 +86,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -97,7 +97,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -963,7 +963,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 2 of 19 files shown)


samsoir:
Follow-up filed as #216 for the `chunk_disk_bytes_written` tier conflation noted in the PR description. Deliberately out of scope here — fixing it changes TUI throughput numbers.

PR #220, opened 2026-08-09
fix(config): accept decimal sizes and round the memory recommendation

### LINKED ISSUE #218: [Bug]: Automatic setup creates fractional values
### Bug Description

One of the values, I think it's the memory can generate floating point values for the config.ini e.g. 2.6 G in my particular case (which is 32/12) but when the main app is run this is an invalid value 

Easy fix : correct in config.init to an integer, it seems to work fine.

Platform OpenSUSE Tumbleweed , Rust from RustUp install. 

### Steps to Reproduce

Run the setup
Accept the default for my 32Mb RAM
Install the packages
Try to run it

### Expected Behavior

It either should allocate 2.6G of RAM or the setup should set it to 2 or 3.

### Actual Behavior

Set to 2.6 (32/12) which caused an error

### System Diagnostics

```text
(not necessary)
```

### Relevant Logs

```text

```

### Screenshots

_No response_

### Severity

Minor functionality broken

### Checklist

samsoir:
Fixes #218

> **Note on auto-close:** GitHub only acts on a closing keyword when a PR merges into the *default* branch (`main`). This PR targets `develop/0.4.7`, so the keyword above will not fire on merge here — `Fixes #218` is also in the final commit's message, which closes the issue automatically when `develop/0.4.7` reaches `main`.

## Summary

@x16-dev reported that the setup wizard wrote `memory_size = 2.6 GB` into `config.ini` on a 32 GB machine, and the application then refused to start. Workaround was hand-editing it to a whole number.

The wizard turned out not to be the culprit. **`format_size` and `parse_size` are inverse functions living in the same file that do not round-trip:**

- `format_size` emits one decimal place for non-exact multiples — its own doctest asserts `format_size(1536 * 1024 * 1024) == "1.5 GB"`.
- `parse_size` did `num_str.parse::<usize>()`, rejecting any decimal outright.

So the application could not read back a config it had written. `recommended_memory_cache` returns exact bytes (`RAM / 12`); the fraction appeared only at write time. A 32 GB machine reports ~31.2 GiB of RAM, so `RAM / 12` lands on 2.6 GiB — reliably fractional, which is why this reproduced consistently for the reporter and not for everyone.

This affected more than the wizard: `config get cache.memory_size` piped into `config set` could not round-trip its own output either.

## Changes

**`parse_size` accepts decimals** (`config/size.rs`) — the load-bearing fix. Any value `format_size` can emit now reads back, including hand-edited ones.

Three guards are load-bearing and deliberate. Rust's float→int `as` casts *saturate* rather than trapping, so a naive `f64` rewrite would turn `"-1GB"` into `Ok(0)`, `"infGB"` into `Ok(usize::MAX)`, and an overflow into a saturated value where `checked_mul` previously errored — trading one loud failure for three silent ones. The parser rejects non-finite and negative values explicitly and bounds-checks before the cast.

A bare-`B` suffix branch was also needed: `format_size` emits `"512 B"` for sub-KB values, a second instance of the same round-trip failure that the new property test exposed. It is matched *after* GB/MB/KB, since `"2GB"` also ends in `B`.

**`recommended_memory_cache` rounds to the nearest whole GB** (`system/recommendations.rs`) — polish, not the fix. It mirrors `recommended_disk_cache`, which already floors to a round step "so the value is round and predictable in the wizard".

| System RAM | Before | After |
|---|---|---|
| 4 GB | 500 MB (floor) | 500 MB (floor) |
| 8 GB | ~683 MB | 1 GB |
| 16 GB | ~1.3 GB | 1 GB |
| 32 GB | ~2.7 GB | 3 GB |
| 64 GB | ~5.3 GB | 5 GB |
| 128 GB | ~10.7 GB | 11 GB |

*Nearest*, not floor: flooring would drop an 8 GB machine to the 500 MB minimum, whereas rounding lifts it to a clean 1 GB, still well under the `RAM/4` ceiling.

**Documentation corrected** — the wizard's own `"RAM ÷ 12"` hint (the screen the reporter was looking at, which showed both numbers and so was visibly not division once rounded), `docs/configuration.md`, and two doc comments that contradicted the tables directly beside them.

## Related Issues

Fixes #218

## Test Plan

- [x] `make pre-commit` passes — fmt, clippy `-D warnings`, full suite, doctests
- [x] **Round-trip property test** — `parse_size(&format_size(n))` recovers `n` across exact GB, fractional GB, fractional MB and sub-KB values. This is the property whose absence caused the bug
- [x] Three rejection cases covered: negative, non-finite, overflow — each must error rather than saturate
- [x] `parse_size("2TB")` still errors; the long-standing test is unchanged
- [x] `format_size` output unchanged — only the parser widened
- [x] Recommendation table above covered by a table-driven test with hand-computed expected values
- [x] Two pre-existing tests (`memory_cache_floored_at_500mb_for_tiny_systems`, `memory_cache_capped_at_quarter_ram`) pass untouched — the latter is the ceiling coverage
- [x] **End to end on this machine** (91.9 GB RAM): `recommended_memory_cache_display()` returns `"8 GB"` — a whole number, no decimal point

## Checklist

- [x] Documentation updated if applicable

## Notes for the reviewer

**The overflow guard's `>` is correct and must not be tightened to `>=`.** `usize::MAX as f64` rounds *up* to 2^64, and `format_size(usize::MAX)` emits `"17179869184 GB"` which multiplies back to exactly 2^64 — so `>=` would reject the formatter's own output and break the round-trip property at the top of the range. There is a comment at the site saying so.

**Six sites encoded the old unrounded contract**, including a doctest that runs in `make pre-commit` and an existing test whose *name* stated the formula. All updated; the two tests that should survive were verified untouched.

**Behaviour change worth noting:** `"10B"` now parses where it previously errored. Intentional — it is what `format_size` emits for sub-KB values.

**Out of scope, noted:** `publisher/config.rs` contains a second, independent `parse_size` returning `u64` with divergent semantics. Not reachable from `format_size` and not touched here, but a consolidation candidate.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -14,7 +14,7 @@ The wizard auto-detects your X-Plane installation, system hardware (CPU, memory,
 
 1. **X-Plane Custom Scenery** — auto-detected from your X-Plane install, with fallback to manual entry.
 2. **Package Location** — where regional scenery packages live on disk.
-3. **Cache Configuration** — cache directory, disk cache size (defaults to 25% of free space, floored to 10 GB), DDS-to-chunk disk ratio, memory cache size (defaults to RAM ÷ 12, clamped to 500 MB – RAM ÷ 4), and disk I/O profile (NVMe / SSD / HDD / auto).
+3. **Cache Configuration** — cache directory, disk cache size (defaults to 25% of free space, floored to 10 GB), DDS-to-chunk disk ratio, memory cache size (defaults to RAM ÷ 12, rounded to the nearest whole GB, clamped to 500 MB – RAM ÷ 4), and disk I/O profile (NVMe / SSD / HDD / auto).
 4. **DDS Encoding** — picks ISPC (CPU) by default, or offers to offload encoding to a secondary GPU when **two or more GPU adapters** are detected. The wizard warns against picking the GPU X-Plane renders on. Single-GPU systems skip the choice and stay on ISPC.
 
 GPU enumeration can take 10–30 seconds on multi-adapter systems while drivers are probed; a spinner shows progress.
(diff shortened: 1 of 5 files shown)


samsoir:
@x16-dev fyi

Issue #157, opened 2026-04-08
bug: prefetch filter skips tiles never generated — coverage gap at sliding box edges


samsoir:
## Summary

The prefetch filter pipeline marks tiles as "cached" (`cache_skipped`) that were never actually generated, creating coverage gaps at the edges of the sliding box. When X-Plane's scenery window reaches these gaps, tiles must be generated on-demand from network, causing sim freezes.

## Root Cause (Confirmed)

**Two-phase region commit creates a dead state** in the BoundaryStrategy:

1. Region enters sliding box → tiles submitted → region marked `InProgress` → tiles added to `cached_tiles` HashSet
2. Sliding box moves forward → `evict_cached_tiles_outside_retained()` removes tiles from `cached_tiles` (outside retention window)
3. Region never promotes to `Prefetched` because `cached_tiles` entries were evicted before promotion could verify them
4. Region is permanently stuck `InProgress` in GeoIndex
5. `PrefetchedRegion::should_prefetch()` sees `InProgress` → blocks the region from being retried
6. Tiles are never generated — region is in a dead state

**Evidence from flight test log:**
- Region `+40-092` marked `InProgress` repeatedly (every ~2 min from 23:41 to 00:00+)
- **Zero `Prefetched` promotions** for any region in the entire session
- `cache_skipped=200/200` in filter summary — all planned tiles excluded at region level
- 687 FUSE on-demand jobs created when X-Plane needed the tiles

**Key code locations:**
- Region promotion: `prefetch/adaptive/boundary_strategy.rs:104-130`
- Tile eviction: `prefetch/adaptive/boundary_strategy.rs:192-219`
- Region filtering: `prefetch/adaptive/prefetch_box.rs:206-210`
- `cached_tiles` HashSet: `prefetch/adaptive/coordinator/core.rs:128`

## Approved Fix: Approach C — Reset stale InProgress regions with protections

When an `InProgress` region's tiles have been evicted from `cached_tiles`, evaluate whether to promote or demote:

### Protection 1: Check disk before demoting (handles the common case)
Before demoting an `InProgress` region, check if the tiles actually exist on DDS disk cache. If they do, the tiles WERE generated — `cached_tiles` just lost track. Promote to `Prefetched` instead of demoting.

### Protection 2: Minimum InProgress duration
Don't evaluate a region until it's been `InProgress` for at least the stale timeout (120s default). This prevents premature demotion while jobs are still in-flight.

[...]

PR #223, opened 2026-08-12
fix(prefetch): one region tile-set definition for submit, promote and rescue (#176)

### LINKED ISSUE #176: prefetch: promotion correctness is downstream of scenery index completeness

samsoir:
## Summary

> **⚠️ This branch is NOT merge-ready on green CI alone.** It changes core prefetch promotion logic and requires live X-Plane 12 validation over multiple 500nm+ flights before sign-off. The PR is open now as a review and testing surface — please fly it. Acceptance criteria are in the Test Plan below.

Region promotion decided whether a DSF region was fully cached by asking which tiles the region *should* contain. There were **two implementations of that question**, and the promotion path was running the wrong one.

Both were born identical in `69658ca` (#58). Twelve hours later `ef47aa1` added a region filter and deduplication — and touched **one file**:

```
 .../src/prefetch/adaptive/coordinator/core.rs | 268 ++++++
 1 file changed
```

The `BoundaryStrategy` copy never got the fix, and `88fde88` (#157) later made it `pub` so the promotion and rescue paths could both call it. **Promotion has been running the pre-fix copy of a function that was fixed five months ago.**

### Three defects, none of which needed an incomplete scenery index

**1. Promotion demanded tiles from neighbouring regions.** A 45nm radius circle around a 1° region centre extends ~15nm past its north and south edges. `promote_completed_regions` required *all* of them on disk — including tiles belonging to adjacent regions that were never submitted for this one. Under a moving prefetch box the leading edge never satisfies that, so the fast path effectively could not fire.

**2. The rescue path promoted on a one-tile sample.** `tiles.first()` — one arbitrary tile, from unordered iteration, deciding a whole region.

Together these mechanically explain #172's observed **61:4 rescue-to-normal promotion ratio**. The fast path was gated on a condition it could not meet, so 61 of 65 promotions were decided by a single-tile sample. #172 attributed that ratio to the `cached_tiles` shadow and fixed it; this cause was never addressed.

Note the composition hazard: each defect looks *conservative* alone ("check more tiles", "keep a safety net"). Composed, they produce the least safe outcome available — the strict check never fires, so the loose one decides everything.

**3. The geometric fallback was a 16-point sample.** `expand_to_tiles(region, 14)` sampled a 4×4 grid of a region that really holds a median of **156** tiles (p95 352, max 1050, measured on a real 11-package install). "All tiles cached" meant "all 16 of 156".

### The original hypothesis was measurably false

#176 was filed on the theory that `.ter` parse failures leave the index incomplete. On a full install the index holds **4,453,034 tiles**, and the eleven independent `read_dir` counts in the cache header sum to exactly 4,453,034 — **parse success is 100% across 4.45M files**. That failure mode has zero occurrences on the only install we can measure.

The observability half of this PR is therefore not bug-hunting. Its job is to keep a verified fact verified, and to make a future malformed package visible.

## Changes

**`SceneryIndex::tiles_in_region` is now the single definition** of a region's tile set, consumed by the submit, promote and rescue paths alike. A `DsfRegion` *is* a grid cell — `cell_key` floors by `cell_size` (default 1.0) and `DsfRegion` floors by 1.0, the same partition — so the default path is one `HashMap` lookup, replacing a 16-key (4×4) cell scan plus per-tile distance maths plus a filter that together recovered exactly the contents of one cell. (Nine is the number of cells a 45nm circle geometrically intersects at mid-latitudes; the deleted query's own `cells_searched` debug field computed sixteen — its `.ceil()` on the upper bound over-scanned a full ring beyond the circle.)

**The rescue path now uses the same full-coverage predicate as the fast path.** The single-tile sample is gone.

**The geometric fallback is deleted.** A region with no indexed tiles is marked `NoCoverage` rather than prefetched from a 16-point guess. This also revives `mark_no_coverage`, previously unreachable below 85° latitude.

**Four dead functions removed** — `tiles_for_region`, `expand_to_tiles`, `land_tiles_near`, `tiles_near_at_zoom`. All were `pub` in a library crate, which is why `clippy -D warnings` never flagged them: rustc treats any `pub` item in a lib as potentially used downstream, and this workspace has no downstream consumer. That is the same mechanism that hid the drifted duplicate after #157 made it `pub`.

**`.ter` parse failures are counted and surfaced.** Per-file logging stays at `debug!` — at 4.45M files a wholesale-malformed package logging per-file at `warn!` would emit millions of lines, the exact flood #209 removed. The visible signal is one aggregated `warn!` per package with a count and up to five sample paths. `scenery-index status` gains a completeness line; on a real install it reads `Indexed: 4453034 of 4453034 .ter files (100.0%)`.

**A 60s `Prefetch sample` line**, on the same cadence as #209's memory sample:

```
Prefetch sample uptime_s=… regions_in_progress=… regions_prefetched=…
                regions_nocoverage=… promotions_normal=… promotions_rescue=…
                state_diverged=… regions_demoted=… fuse_generations=…
```

The per-cycle `debug!` distribution line is unchanged. This exists because **all five flight-test criteria below are now read from this one line**: region state for criteria 3 and 4, previously visible only at `--debug`; the promotion counters for criterion 1; and, since remediation, `fuse_generations` (mirroring the existing `fuse_jobs_submitted` counter) for criterion 5 — previously unreadable at any log level, since per-tile FUSE request logging is `debug!`-only (#209). `--debug` is impractical to run for a whole 500nm flight and is overwritten on the next launch.

**`PrefetchStateObserver`** — when FUSE has to *generate* a tile in a region prefetch marked `Prefetched` or `NoCoverage`, the claim was wrong. Three causes produce that (an index gap, a premature promotion, post-promotion eviction) and the response is identical for all three: clear the state so the region is re-prefetched. This runs synchronously inside `do_request()` (`fuse3/shared.rs`) — `self.on_dds_response(...)` fires immediately before `response.data` is returned — so it is *on* the texture-response path, not off it. It cannot corrupt the served DDS bytes (`response.data` is set before this call and unaffected by it), and the one lock it can take — the demotion rate-limit mutex — is reached only on the divergence branch, after two early returns, so contention there is not a practical concern. Demotion is rate-limited to once per region per 120s — eviction inside the retained window recurs, and an unbounded demote → re-prefetch → evict loop would churn on a long flight — while **every** divergence is counted, so the two numbers can diverge.

It correctly stays silent for `InProgress` regions (expected to miss) and for cache hits (the system working).

## Related Issues

Fixes #176

> **Note on auto-close:** GitHub only acts on a closing keyword when a PR merges into the *default* branch (`main`). This PR targets `develop/0.4.7`, so the keyword above will not fire on merge here — `Fixes #176` is also in commit `8541ee7`'s message, which closes the issue automatically when `develop/0.4.7` reaches `main`.

## Changes since review

Fifteen commits landed after the adversarial review (`3d73a63..f253bb1`), closing gaps the review found in the promotion logic itself, not just in its tests and docs.

**One completeness predicate now serves both promotion paths.** `BoundaryStrategy::region_disk_state` (`e1ce72c`) replaces the two independently-drifted copies that used to live in `promote_completed_regions` (fast path) and `evaluate_stale_regions` (rescue path) — the same copy-drift shape #176 was filed to remove, just one level down. It returns a four-state `RegionDiskState` (`Complete` / `Incomplete` / `NoTiles` / `Unknown`) instead of a `bool`. `Unknown` — no `DdsDiskCacheChecker` and/or no `SceneryIndex` wired — is the important addition: previously the rescue path's `_ => false` treated "no authoritative source to consult" identically to "checked and it's absent," so an unconfigured disk checker could strike a region toward permanent `NoCoverage`. `Unknown` is now left untouched and does not consume a retry. This incidentally fixed a second, previously undocumented copy-drift case: "scenery index absent but a checker present" used to strike toward permanent `NoCoverage` on the rescue path while the fast path treated the identical situation as a harmless no-op (an early return, no side effects).

**Region coverage is now a disjunction, not just the XEL cache.** `cc85bc0` adds a `tile_is_covered` check inside `region_disk_state`: a tile counts as covered if it's in the XEL DDS disk cache **or** ships as a real `.dds` file inside an installed scenery package (`OrthoUnionIndex::dds_tile_exists`, chunk-coordinate lookup). Prefetch's stage-3 filter already skips tiles that exist in an installed package — it never writes them to the XEL cache — so without this, a region entirely covered by package imagery could never be confirmed by promotion and would ratchet to `NoCoverage` over land after `MAX_REGION_ATTEMPTS`. That's exactly the false alarm flight criterion 4 below is watching for.

**Two new cumulative counters.** `97db4c7` adds `promotions_normal` and `promotions_rescue` to the `Prefetch sample` log line above, wired through the same event → client → state → daemon chain as the existing divergence/demotion counters. This is what makes flight criterion 1 measurable at default log level for the first time — see the Test Plan.

**Region strike counts have three exits, and all three now clear or preserve the count correctly (`1e209a3`, `fd89b48`).** `region_attempts` previously only ever incremented. Promotion clears the count on both paths (`1e209a3`) — the strike belongs to the failed episode, not the region, so a region that failed twice, recovered, and was later demoted by `PrefetchStateObserver` shouldn't be retired on one fresh failure using strikes from hours earlier. The retry branch deliberately does **not** clear it: that branch removes the region's `GeoIndex` entry to make it eligible for re-prefetch, and `region_attempts` is the only thing left remembering the region failed last time; clearing it there would make `MAX_REGION_ATTEMPTS` unreachable. The third exit — **retirement** to `NoCoverage` — was missing this entirely until `fd89b48`: a region retired at `MAX_REGION_ATTEMPTS` strikes kept that count, so when `PrefetchStateObserver` later demoted it out of `NoCoverage` (correctly, because it observed the claim was wrong), the region re-entered the prefetch cycle already at the strike ceiling — one single fresh failure re-hit `>= MAX_REGION_ATTEMPTS` and instantly re-retired it, with zero real retries. Because strike-driven retirement only ever fires on regions with indexed tiles (ocean regions take the empty-index path and never touch the counter), every region this hit was, by construction, land — biasing flight criterion 4 toward false alarm. `fd89b48` clears the count on retirement too, so a demoted region gets a fresh set of `MAX_REGION_ATTEMPTS` strikes each time; the observer's own demotion rate limit (one per region per ~120s) bounds how often that can recur, not this counter. The fix also resolves a self-contradictory warning: the retirement log line's fixed text ("after 3 failed attempts") could previously be paired with `attempts=4` from a leftover count — now the two agree.

**Four test gaps closed, each by a test confirmed to fail against the specific break it targets.** `c3b81e9` closes two: the planning-path `NoCoverage` marking had no test that ran a real planning cycle and read `GeoIndex` state — deleting the marking made the new test's submission-count assertion fail; and the two-phase `InProgress`→`Prefetched` commit boundary was unenforced because an existing test's mock played both the filter authority and the promotion authority — replacing the disk-verified promotion with a direct `Prefetched` insert left that test green, which the new test (emptying the plan via the memory-cache filter while the disk checker reports every tile absent) catches. `8da109d` closes two more: `with_metrics_client` had zero test callers, leaving the region-state gauge classification unverified — deleting the `metrics.prefetch_region_state(...)` emission, swapping its argument order, and swapping the `is_in_progress`/`is_prefetched` fold arms each independently made the new test fail (the latter two both producing `expected PrefetchRegionState { in_progress: 1, prefetched: 2, no_coverage: 3 }, got PrefetchRegionState { in_progress: 2, prefetched: 1, no_coverage: 3 }`); and a zoom-mismatch regression test had separately (accidentally) mismatched row/col against the real region tiles, so a checker mutated to ignore the zoom argument entirely would still have failed to promote for the wrong reason — with the fixture corrected to the region's actual tiles (row 1373–1386), that same zoom-blind mutation now promotes and the test correctly catches it.

**Dead code removed.** The `tiles_near` radius query and its `approximate_distance_nm` helper (`8cffa8b`) — the deleted duplicate `tiles_for_region` was literally `tiles_near(centre, 45.0)` mapped to tile coordinates, so leaving `tiles_near` exported kept that drift-prone pattern one call away from returning. And the write-only `cached_tiles` `HashSet` shadow (`bba4a9d`), confirmed write-only before deletion — every remaining use was an insert/clear/retain, or a `.len()` read only for logging and tracing, never for control flow.

**Criterion 5 now has a field.** The final whole-branch review found that the `Prefetch sample` line documented itself as sufficient for "the #176 acceptance criteria" while actually carrying fields for criteria 1–4 only — criterion 5 (on-demand FUSE generations falling during cruise) had nothing to read at default log level, since per-tile FUSE logging was demoted to `debug!` in `ea02c13`. `480cdf9` emits the existing `fuse_jobs_submitted` counter (already aggregated in `AggregatedState`, previously surfaced only to the TUI) as `fuse_generations` on the same line. All five flight criteria now come from one line.

Also landed: a `TempDir` leak in a new test helper, fixed by borrowing instead of leaking (`701a567`); the FUSE-side demotion interval and the coordinator's `stale_region_timeout` default now derive from one canonical constant instead of a comment-documented mirror (`1c473bf`); the design docs (`cea1575`, `b7bdfb6`) were brought up to date with all of the above — including the corrected four-outcome `evaluate_stale_regions` description, the `GeoIndex` layer inventory (`PrefetchedRegion`/`RetainedRegion` were missing), and a retry-count off-by-one (a region gets at most 2 revert-to-absent retries before the third strike retires it, not 3); and, from the final review's cleanup pass (`f253bb1`), the orphaned `AdaptivePrefetchCoordinator::reset()` was deleted — no callers anywhere, and its doc comment's claim to reset state for "teleporting or starting a new flight" had already gone false when `bba4a9d` removed its only substantive statement — alongside three more stale documentation sites, including a `BoundaryStrategy` "Interface" block that showed a fabricated trait impl and a `calculate_prefetch` signature naming the deleted `cached_tiles` shadow.

### Automated

- [x] `make pre-commit` passes — fmt, clippy `-D warnings`, 2474 passing lib tests (9 ignored), 63 doctests
- [x] **Both regression tests failed before the fix**, with the predicted assertion mismatches. This is the evidence the defect was real rather than a misreading — everything asserted about it came from reading code and `git log`, not from observing a failure
- [x] Promotion is not blocked by a missing tile in an adjacent region (defect 1)
- [x] The rescue path requires full coverage, not one tile (defect 2)
- [x] A region with no indexed tiles is marked `NoCoverage` and submits nothing
- [x] `tiles_in_region` correct for negative coordinates, non-default `grid_cell_size`, and deduplication; sea tiles remain in scope
- [x] The FUSE wiring chain is covered end-to-end (`request_dds_impl → do_request → on_dds_response → observe → GeoIndex`), mutation-verified
- [x] Metric emission asserted over a real channel, so a swapped or deleted emission is caught

### Empirical validation against a real install

- [x] **Region assignment verified across all 4,453,034 cached tiles**: `floor(LOAD_CENTER)` and `floor(TileCoord::to_lat_lon())` produce the same `DsfRegion` for every tile — **0 mismatches**. This is an empirical result, not a structural one: `tiles_in_region` (f32 `.ter` coordinates) and `PrefetchStateObserver` (f64 Mercator centre) are two independent derivations with no code-level link between them, and measurement is the only evidence they agree. It's a strong result — 0 mismatches across the full install — but a future edge case (e.g. a tile centre that lands exactly on a degree boundary under floating-point rounding) isn't ruled out by construction
- [x] Tiles per region measured: p50 156, p95 352, max 1050. Latitude range −56.0…83.7, safely inside `MAX_LAT`

### Flight testing — REQUIRED before merge

Multiple flights of 500nm or greater in X-Plane 12. Measured per flight from the `Prefetch sample` line at default log level:

- [ ] **1. Normal-path promotions dominate rescue-path promotions.** Now directly measurable: compare `promotions_normal` against `promotions_rescue` in the `Prefetch sample` line (both cumulative totals for the flight, like the other counters on that line). Previously this criterion had no counter to read at any log level — it could not actually be judged from a flight log before remediation. Baseline from the #172 LOWW log: 4 normal vs 61 rescue (6% normal). Post-fix `promotions_normal` must be the majority
- [ ] **2. `state_diverged` stays at or near zero.** Non-zero is not automatically a failure — post-promotion cache eviction produces it legitimately — but a cluster in one area needs explaining before sign-off. **This is a cumulative total for the whole flight, not a per-60s rate** (`AggregatedState::reset()` has no production caller)
- [ ] **3. `regions_in_progress` does not grow monotonically.** A ratchet means regions enter `InProgress` and never leave — the failure mode defect 1 produces today
- [ ] **4. `regions_nocoverage` corresponds to ocean.** The riskiest consequence of dropping the geometric fallback: a count rising over land means the empty-cell-means-ocean assumption is wrong. Cross-check against the route
- [ ] **5. On-demand FUSE generations fall during cruise.** Now read from `fuse_generations` on the same `Prefetch sample` line as the others (cumulative since process start), rather than an unmeasured comparison against a pre-change baseline. The user-visible outcome; the others are mechanism

**The signal that something is wrong:** `regions_nocoverage` climbing alongside `Region marked NoCoverage after 3 failed attempts` warnings.

## Checklist

- [x] Documentation updated (`CLAUDE.md`, `docs/dev/adaptive-prefetch-design.md`, `docs/dev/geo-index-design.md`, `docs/dev/memory-telemetry.md`)

## Notes for the reviewer

**A pre-existing bug this fixes, but not for every patch region.** Patched regions previously received 16 geometric tiles that `filter_patched_regions` discarded — and because the region was then never fully submitted, it was re-planned *every single cycle, forever*. Deleting the geometric fallback fixes this **only when the scenery index has no tiles for the region at all**: `get_tiles_for_region` now marks such a region `NoCoverage` directly instead of manufacturing a 16-tile guess, so it stops re-entering the plan. Patch regions that overlay package coverage — the common case, since that's usually why a patch exists in the first place — aren't covered by that fix: the scenery index still returns real tiles for them, `filter_patched_regions` still discards all of them, and (see below) `mark_fully_filtered_regions` deliberately excludes patch-owned regions from being marked `InProgress`. Such regions keep being re-planned and re-filtered every cycle, exactly as before this PR — the comment on `mark_fully_filtered_regions` in `core.rs` says so directly.

**A second one, found and fixed during final review.** `execute()` is skipped when the plan is empty and `mark_in_progress` only fires inside it, so a region whose planned tiles all filtered out as already-cached was never marked anything — re-planning every 2s indefinitely and making `regions_prefetched` under-report. Pre-existing, but this branch creates a routine path into it via demotion, and an under-reporting gauge would have corrupted criteria 1 and 3 above. `mark_fully_filtered_regions` now marks such regions `InProgress` and lets the disk-verified `promote_completed_regions` confirm them — nothing reaches `Prefetched` without the full-coverage check.

**Patch-owned regions are deliberately excluded from that marking.** Prefetch never puts a patch region's tiles on the XEL DDS disk cache, so promotion could never confirm the claim, and every patch-owned region would reach `NoCoverage` after ~6 minutes with a misleading "failed attempts" warning in the flight log. `PatchCoverage` is a whole-region `GeoIndex` layer, so the exclusion is exact rather than heuristic.

**Coverage, stated honestly.** Measured lib-only at **61.49% on this branch vs 61.12% at the base commit** — the branch improves it, and `state_observer.rs` lands at 27/28 lines. This was measured at `3d73a63`, before the fifteen remediation commits that followed review (which both added tests and deleted code — either can move the number, and deleting well-covered dead code can move it *down* while still improving the codebase); it has not been re-measured since, and this note doesn't claim otherwise. Regardless of the exact current figure, the repository does not meet its own `COVERAGE_MIN=80` and **did not before this branch**, so please don't read the number as a regression. Related: `make coverage-check` prints "Coverage below 80%!" on *any* non-zero tarpaulin exit, including a failed test — it reported that twice during this work without ever computing a percentage. Worth its own chore issue.

**Six tests fail under `cargo tarpaulin`** instrumentation (executor daemon coalescing, `storage_limiter::test_with_defaults`, `test_telemetry_timeout`). Also checked at `3d73a63`, not re-checked since. Verified pre-existing at that point — the same failures occurred on the unmodified base commit. They pass normally in 0.15s.

**No `CHANGELOG.md` entry**, per this repo's convention of compiling the changelog at release-cut time from merged PRs. An empty `[Unreleased]` section is correct here, not an oversight.

**Known limitation worth stating:** when the cause of a divergence is a genuine scenery-index gap, the observer can only *flag* it, never repair it — the index is immutable after build, so the next cycle re-derives the same empty set. The loop is bounded (rate limit plus the `should_prefetch` guard), but don't expect self-healing for that specific cause.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,7 +135,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
     - `PrefetchStateObserver` - FUSE-side divergence detector; demotes a `Prefetched`/`NoCoverage` region back to *absent* when FUSE has to generate a tile on demand there, rate-limited to one demotion per region per 120s window
     - Submits jobs to shared job executor daemon via `DdsClient` trait
     - Mode selection: Aggressive (>30 tiles/sec), Opportunistic (10-30), or Disabled
-    - **Four-tier filtering**: Local tracking → Memory cache → Patched region exclusion (via `GeoIndex`) → Disk existence (via `OrthoUnionIndex`)
+    - **Four-tier filtering**: Memory cache → Patched region exclusion (via `GeoIndex`) → Installed package disk (via `OrthoUnionIndex`) → XEL DDS disk cache (the write-only `cached_tiles` local-tracking shadow tier was deleted in #176 after it was confirmed to never be read)
     - **Two-phase region commit**: Regions marked `InProgress` in GeoIndex during prefetch, promoted to `Prefetched` on completion
     - **Stale telemetry safe mode**: pauses tile submissions when telemetry stale >5s; on resume, reads `on_ground` from AircraftState to reset PhaseDetector before next cycle
     - See `docs/dev/adaptive-prefetch-design.md` for design details
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,6 +132,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
     - `SimState` - Direct sim state detection via X-Plane Web API (paused, on_ground, scenery_loading, replay)
     - `TransitionThrottle` - Grace period + ramp-up after Ground-to-Cruise transition
     - `LoopState` - Per-cycle runner state including `telemetry_paused` flag for stale telemetry safe mode
+    - `PrefetchStateObserver` - FUSE-side divergence detector; demotes a `Prefetched`/`NoCoverage` region back to *absent* when FUSE has to generate a tile on demand there, rate-limited to one demotion per region per 120s window
     - Submits jobs to shared job executor daemon via `DdsClient` trait
     - Mode selection: Aggressive (>30 tiles/sec), Opportunistic (10-30), or Disabled
     - **Four-tier filtering**: Local tracking → Memory cache → Patched region exclusion (via `GeoIndex`) → Disk existence (via `OrthoUnionIndex`)
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -673,7 +673,7 @@ For each BoundaryCrossing:
        knows nothing about yields no tiles and is marked NoCoverage
 
     5. Apply four-tier filter:
-       Local tracking → Memory cache → Patch exclusion → Disk existence
+       Memory cache → Patch exclusion → Installed package disk → XEL DDS disk cache
 
     6. Order tiles:
        Primary: urgency rank (depth 0 before depth 1 before depth 2)
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -248,7 +248,18 @@ Each DSF region in the XEL Window has one of four states:
 This follows a **two-phase commit** pattern:
 1. **Reserve**: Region marked `InProgress` when tiles are submitted (prevents duplicate bulk submissions)
 2. **Confirm**: Region marked `Prefetched` when tiles are confirmed in cache
-3. **Timeout**: If `InProgress` exceeds `stale_region_timeout`, reverts to *absent* for re-evaluation
+3. **Timeout**: If `InProgress` exceeds `stale_region_timeout`, `evaluate_stale_regions`
+   consults `BoundaryStrategy::region_disk_state` — the same completeness predicate the
+   fast path uses — and picks one of four outcomes:
+   - Disk shows every tile present → promote straight to `Prefetched` (the rescue path;
+     see [Region Tile-Set SSOT](#region-tile-set-ssot) below)
+   - Disk shows tiles missing, under `MAX_REGION_ATTEMPTS` (3) → reverts to *absent* for
+     re-evaluation next cycle, incrementing the region's strike count
+   - Disk shows tiles missing, at `MAX_REGION_ATTEMPTS` → retired to `NoCoverage`
+     **permanently for the session** — this is the event flight-test criterion 4 keys on
+   - No authoritative source to consult (no `DdsDiskCacheChecker` and/or no
+     `SceneryIndex`) → left `InProgress` untouched; this does **not** count as a retry,
+     since "cannot tell" must not be treated the same as "checked and it is absent"
 
 A region can reach step 1 without submitting anything. When the filter
 pipeline removes a region's *entire* planned tile set as already cached, there
@@ -272,6 +283,48 @@ whether a region is complete. Earlier code reconstructed the tile set from a
 45nm radius query per path, which returns a circle overlapping neighbouring
 regions rather than the region itself — the source of several promotion bugs
 fixed under #176.
+
+`BoundaryStrategy::region_disk_state` (`xearthlayer/src/prefetch/adaptive/boundary_strategy.rs`)
+is the single completeness predicate built on top of that tile set, consumed
+by both the fast path (`promote_completed_regions`) and the rescue path
+(`evaluate_stale_regions`). It returns one of four states — `Complete`,
+`Incomplete`, `NoTiles`, or `Unknown` — rather than a `bool`, because the
+answer feeds a decision whose terminal outcome is permanent exclusion
+(`NoCoverage`): "we cannot tell" (`Unknown` — no `DdsDiskCacheChecker` and/or
+no `SceneryIndex` wired) must stay distinguishable from "we checked and it is
+absent" (`Incomplete`/`NoTiles`), or a region could be retired to `NoCoverage`
+on a cycle where it was never actually checked.
+
+**Coverage is a disjunction.** A tile counts as covered if it is present in
+the XEL DDS disk cache (`DdsDiskCacheChecker`) **or** ships inside an
+installed scenery package (`OrthoUnionIndex::dds_tile_exists`) — prefetch
+deliberately never downloads tiles the disk filter (stage 3 of the four-tier
+filter) already excluded as package-shipped, so promotion must recognise
+that same evidence or a region covered entirely by package imagery could
+never be confirmed and would ratchet to `NoCoverage` over land. The two
+lookups use different coordinate systems: the XEL DDS cache is keyed by
+**tile** coordinates (`tile.row, tile.col, tile.zoom`), while
+`OrthoUnionIndex::dds_tile_exists` is keyed by **chunk** coordinates
+(`tile.chunk_origin()`) — mixing them up silently returns "not covered" for
+every tile (the #172 keying bug).
+
+**Strike counts survive the retry branch, not the promote branch.** A
+region's `region_attempts` strike count is cleared on promotion (both the
+fast path and the rescue path) but deliberately left untouched on the retry
+branch of `evaluate_stale_regions`. This looks backwards until you notice
+what the retry branch does: it *removes* the region's `PrefetchedRegion`
+entry from the `GeoIndex` so the region is eligible for re-prefetch — and
+that removal is exactly what makes `region_attempts` (a separate
+`HashMap<DsfRegion, u8>` on the coordinator, not stored in the `GeoIndex`)
+the only thing that still remembers the region failed last time. Clearing it
+on retry, by naive symmetry with the promote branch, would make
+`MAX_REGION_ATTEMPTS` unreachable: every retry would restart the count at
+zero and a region that never succeeds would retry forever instead of
+eventually being retired to `NoCoverage`. Promotion clears it for the
+opposite reason — a promoted region's failure history belongs to the episode
+that failed, not to the region, so a region that recovers and is later
+demoted by `PrefetchStateObserver` should not be retired to `NoCoverage`
+after a single fresh failure.
 
 ### Demotion: Closing the Loop on Wrong Claims
 
@@ -315,8 +368,17 @@ wire it up and these become totals since the last reset:
 
 ```
 INFO Prefetch sample uptime_s=... regions_in_progress=... regions_prefetched=...
-     regions_nocoverage=... state_diverged=... regions_demoted=...
+     regions_nocoverage=... promotions_normal=... promotions_rescue=...
+     state_diverged=... regions_demoted=...
 ```
+
+`promotions_normal` counts fast-path promotions (`promote_completed_regions`,
+per-cycle region maintenance); `promotions_rescue` counts stale-region
+rescue promotions (`evaluate_stale_regions`, the `RegionDiskState::Complete`
+arm above). A healthy flight should show `promotions_normal` dominating; a
+high `promotions_rescue` ratio means the fast path is stalling and the
+rescue path is carrying the work — see [Region Tile-Set
+SSOT](#region-tile-set-ssot).
 
 `state_diverged` and `regions_demoted` are expected to stay at (or near) zero
 on a healthy flight; a climbing `state_diverged` with a flat `regions_demoted`
@@ -1009,8 +1071,11 @@ SimState (X-Plane Web API)
 
 If some tiles in a submitted region fail:
 - Region stays `InProgress`
-- After `stale_region_timeout` (120s), reverts to *absent*
-- Next evaluation cycle: target diff finds the region, re-submits
+- After `stale_region_timeout` (120s), `region_disk_state` reports `Incomplete` — see
+  [Region States](#region-states-geoindex-prefetchedregion-layer) above — so the region
+  reverts to *absent* for retry, up to `MAX_REGION_ATTEMPTS` (3) times before being
+  retired to `NoCoverage`
+- Next evaluation cycle (while retries remain): target diff finds the region, re-submits
 - Four-tier filter handles individual tile dedup (only failed tiles re-submitted)
 
 ---
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -416,7 +416,6 @@ Regions previously retained but now outside → evicted from GeoIndex
 
 When a region leaves the retained area, the coordinator also evicts:
 - Its `PrefetchedRegion` entry (making it eligible for re-prefetch if the aircraft returns)
-- Any `cached_tiles` entries for tiles in that region (allowing fresh memory cache queries)
 
 `InProgress` regions are never evicted — they represent actively running prefetch jobs.
 
@@ -770,15 +769,21 @@ If backpressure reduces batch size, the most urgent axis is served first.
 
 ### Interface
 
+> **Stale**, like the algorithm block above: `BoundaryStrategy` does not
+> implement `AdaptivePrefetchStrategy` — it exposes static functions
+> (`promote_completed_regions`, `evict_non_retained`, `region_disk_state`)
+> instead. The trait itself still exists (`xearthlayer/src/prefetch/adaptive/strategy.rs`);
+> its current `calculate_prefetch` signature, for reference:
+
 ```rust
-impl AdaptivePrefetchStrategy for BoundaryStrategy {
+pub trait AdaptivePrefetchStrategy: Send + Sync {
     fn calculate_prefetch(
         &self,
         position: (f64, f64),
-        predictions: &[BoundaryCrossing],
-        xel_window: &XelWindow,
-        cached_tiles: &HashSet<TileCoord>,
-    ) -> Option<PrefetchPlan>;
+        track: f64,
+        calibration: &PerformanceCalibration,
+        already_cached: &HashSet<TileCoord>,
+    ) -> PrefetchPlan;
 }
 ```
 
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -250,6 +250,18 @@ This follows a **two-phase commit** pattern:
 2. **Confirm**: Region marked `Prefetched` when tiles are confirmed in cache
 3. **Timeout**: If `InProgress` exceeds `stale_region_timeout`, reverts to *absent* for re-evaluation
 
+A region can reach step 1 without submitting anything. When the filter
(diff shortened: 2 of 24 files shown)


samsoir:
@mmaechtel I'm going to push an alpha build for v0.4.7 once this merges - would appreciate your support testing as the alpha includes your dds mipmap fix as well. There is a small memory leak that I am trying to track down, I will provide more info on what to test once the alpha is shipped.

mmaechtel:
Alpha test report — v0.4.7-alpha.1, Linux x86_64

Provider: Apple Maps · BC1 · GPU compressor (integrated) · `memory_size = 2 GB` · `disk_size = 1500 GB`
Host: 91 GB RAM, NVMe.
Flight: LOWS → EDDF (Salzburg to Frankfurt), A20N, ~73 min.
**Cache was completely empty at start** — everything below is a cold-cache run.

## Memory growth during cruise

Values are taken verbatim from the attached log, one row per 10 minutes:

```
uptime  rss_mb  vm_mb  swap_mb  chunks_ok  dds_disk_mb  chunk_disk_mb  gc_evicted_mb
   0s      41    2410      0            0            0              0        0
 600s    5671   13143      0      575,078       23,818          8,295        0
1200s   12946   17858      0    2,827,116      117,388         37,771        0
1800s   12968   16844      0    2,855,680      118,988         38,125        0   <- idle on ground
2400s   13000   16908      0    2,855,680      118,988         38,125        0   <- idle on ground
3000s   18561   23163    567    4,371,893      181,580         57,710        0   <- swap starts
3600s   25501   30173    225    6,875,881      283,790         96,698        0
4200s   26243   32792   2694    8,292,608      345,529        117,214        0
4380s   26464   32994   2699    8,292,608      345,529        117,214        0   <- end of excerpt
```

`vm_mb` roughly doubled — 16.9 → 32.8 GB — over the 30 minutes between uptime 2400s and 4200s,
while `rss_mb` lagged behind and the host began swapping. Host free memory dropped from ~27 GB
to ~9 GB across that window.

Note the two idle rows at 1800s/2400s: the aircraft was on the ground, nothing was being
downloaded or encoded, and memory sat flat at ~16.9 GB. Growth resumes exactly when tile work
resumes, so it tracks throughput rather than wall-clock time.

Because the cache started empty, this run downloaded and encoded essentially everything along
the route: **8.3M chunks, 345 GB of DDS and 117 GB of raw chunks written to disk in ~70 minutes**,
with `chunks_failed=0` throughout. So it is a fairly aggressive stress case for whatever is
accumulating.

Two observations that may help narrow it down:

- `gc_evicted_mb` is **0** for the entire run. The GC never fired once, even with 462 GB written
  against the 1500 GB budget — so this is growth with the disk GC never having been triggered.
- `chunk_index_entries` tracks `chunks_ok` **exactly** (8,292,608 / 8,292,608) and only ever grows,
  while `mem_cache_mb` stays pinned at 2037 the whole time.

After the flight ended, `vm_mb` settled at a plateau around 34–35 GB rather than continuing to
climb, so I would not call this unbounded linear growth on this evidence. (Those later samples are
not in the attached excerpt — see the note below.) But the doubling under load is clearly in there,
and a cold cache looks like a reliable way to trigger it.

## About the attached log

`lows-eddf-20260816.log` covers startup (09:19:58Z) through 10:33:12Z. I cut it there because
after that point I ran unrelated filesystem scans on the same host, which pushed pages out of
XEarthLayer's RSS into swap and would make the `swap_mb`/`rss_mb` columns misleading. Everything
in the excerpt predates that.

The 20,798 `Failed to submit job - executor may be shutdown` lines are filtered out per the
known-issues list. Everything else is verbatim, including all 27 ERROR lines.

Happy to run this again with `--debug` or a different provider if that would help.

mmaechtel:
[lows-eddf-20260816.log](https://github.com/user-attachments/files/31116069/lows-eddf-20260816.log)

Issue #230, opened 2026-08-17
Rescue-path exact-count scan can burst ~500k stat() calls synchronously on cold-cache terrain entry


samsoir:
The rescue path's exact-count region scan can issue a large synchronous burst of filesystem
`stat()` calls on the prefetch runner, with no yield point, at the moment the aircraft enters
fresh terrain on a cold cache.

Found by the whole-branch review of the #226 fix. **Not yet observed in flight** — filed so the
analysis is not lost and so the acceptance flight knows what to look for. The reviewer's
recommendation was to watch for it rather than fix it late in that branch, and I agreed.

## Mechanism

`BoundaryStrategy::region_disk_state` takes a `CoverageDetail` parameter (added in #226 to keep
the every-2-second promotion path cheap). The rescue path
`AdaptivePrefetchCoordinator::evaluate_stale_regions` passes `ExactCounts`, which scans **every**
tile in the region rather than short-circuiting, because it needs the covered count to tell a
slow region from a stuck one.

For each tile not present in the XEL DDS cache, `tile_is_covered` falls through to
`OrthoUnionIndex::dds_tile_exists` (`ortho_union/index.rs:487`). That is not a map lookup:

- 4 filename prefixes (`ZL`, `BI`, `GO2`, `GO`), each calling `resolve_lazy`;
- `resolve_lazy` always misses its index, because `textures/` is in `LAZY_DIRECTORIES` and is
  deliberately never scanned;
- so each falls into `for source in &self.sources { real_path.exists() }` — a `format!`, a
  `join`, and a `stat()` per installed package, with no caching.

An absent tile therefore costs **4 × N_sources** `stat()` calls. With 11 installed packages,
~44.

## Why it bursts

Regions enter `InProgress` together — the planner marks a batch in one or two cycles as the
aircraft crosses into new terrain. They therefore all cross `stale_region_timeout` (default 120s)
in the **same** maintenance pass. On a cold cache each is stuck with `covered ≈ 0` of ~400 tiles.

Order of magnitude: 30 regions × 400 tiles × 44 stats ≈ **500,000 `stat()` calls in a single
pass**. That pass runs synchronously inside `async fn process_telemetry` with no `.await`, so the
prefetch loop cannot make progress until it completes.

Two aggravating details:

1. #226 also removed a `block_in_place` wrapper from `tile_exists_blocking` — correct in
   isolation, since that half became a plain map lookup, but it was also the only thing telling
[...]

Issue #229, opened 2026-08-17
AggregatedState::reset() has no production caller — counters are cumulative but documented as per-interval


samsoir:
`AggregatedState::reset()` has no production caller. Every field it zeroes is therefore
cumulative since process start, not per-interval — but the code is written, tested and
commented as though the reset runs.

Found during the whole-branch review of #226.

## Evidence

All three call sites are inside test modules:

```
xearthlayer/src/metrics/state.rs:527    (fn test_aggregated_state_reset)
xearthlayer/src/metrics/daemon.rs:1256  (#[cfg(test)])
xearthlayer/src/metrics/daemon.rs:1359  (#[cfg(test)])
```

`grep -rn "\.reset()" xearthlayer/src/metrics/` returns nothing outside `#[cfg(test)]`.

This is consistent with observed behaviour: on a 6h flight, `promotions_normal` reached 238 as a
running total, not a per-interval count.

## Why it is worth fixing rather than documenting

The function is fully exercised by tests, which makes it look live. That is not hypothetical —
during #226 the reset semantics were taken as real from reading `reset()` and its tests, and an
incorrect instruction was issued on that basis. It took an implementer checking the call graph to
catch it. Dead code that tests keep warm is worse than dead code that doesn't, because the tests
are the evidence people reach for.

There is also a real cost to readers of flight logs. The `Prefetch sample` and `Memory sample`
lines are emitted every 60s, and several fields are documented as counters. Because nothing
resets them, judging an acceptance criterion like *"`regions_deferred` should not climb without
bound"* requires **differencing successive samples by hand** rather than reading one line.

## The decision

Either:

1. **Wire it** — call `reset()` after each periodic emission, so the counter fields genuinely mean
   "in the last interval" and a single log line is readable on its own. Gauges are already
   excluded from `reset()`, so the split is in place. This changes the meaning of existing fields
   in flight logs, so it should land at a release boundary and be called out in release notes.
2. **Delete it** — drop `reset()` and its tests, and correct every comment that implies interval
   semantics. Counters stay cumulative; documentation says so plainly, and the log-reading
   guidance says to difference the series.

Option 2 is the smaller change and matches what the code does today. Option 1 is the friendlier
instrument. Worth a deliberate call rather than drift.

Whichever is chosen, the gauge/counter distinction introduced in #226
(`regions_deferred_active` vs `regions_deferred`) must stay intact — that distinction is real in
[...]

Issue #228, opened 2026-08-17
Plan-time NoCoverage marking cannot distinguish 'no scenery index' from 'index says empty'


samsoir:
A second path marks regions `NoCoverage` without ever consulting `region_disk_state`, and it
cannot distinguish "the scenery index says this region has no tiles" from "there is no scenery
index to ask". That is the same `cannot tell ≠ absent` conflation as #226, in a different
function.

Found during the whole-branch review of the #226 fix. **Pre-existing — not introduced by that
branch**, and it does not violate #226's requirement (it never observes `Incomplete`). Filed
separately because it can corrupt #226's acceptance criteria and because the defect class is
identical.

## The path

`AdaptivePrefetchCoordinator::update` marks a region `NoCoverage` whenever the planner finds no
tiles for it — `prefetch/adaptive/coordinator/core.rs:597-603`:

```rust
for region in &new_regions {
    let tiles = self.get_tiles_for_region(region);
    if tiles.is_empty() {
        if let Some(ref geo_index) = self.geo_index {
            self.boundary_strategy.mark_no_coverage(region, geo_index);
        }
    }
```

And `get_tiles_for_region` returns empty when there is no index at all (`core.rs:389-398`):

```rust
match self.scenery_index {
    // An empty result means the index knows of no ortho scenery
    // here. The caller marks the region NoCoverage.
    Some(ref index) => index.tiles_in_region(*region),
    None => Vec::new(),
}
```

The comment states the conflation outright: an empty result is treated as *"the index knows of
no ortho scenery here"*, when it is equally produced by *"there is no index"*.

`scenery_index` is `None` in production whenever the index is empty —
`service/orchestrator/prefetch.rs:143-151`:

```rust
if self.scenery_index.tile_count() > 0 {
    coordinator = coordinator.with_scenery_index(Arc::clone(&self.scenery_index));
} else {
    tracing::warn!("Scenery index is empty (no ortho scenery installed) — prefetch will be inert");
}
```

**"Inert" is not accurate.** With no ortho scenery installed, prefetch does not merely skip work:
on the first cycle it marks **every region in the prefetch box** as `NoCoverage` — over land,
over ocean, without consulting any index — and those regions are then excluded from prefetch for
the rest of the session.

## Why it matters now

1. **It can corrupt #226's acceptance criteria.** One criterion is `regions_nocoverage` non-zero
   only over water. A tester running with no ortho packages installed will see every region
[...]

PR #232, opened 2026-08-18
fix(prefetch): retire only NoTiles regions to NoCoverage (#226)

### LINKED ISSUE #226: Slow regions are retired as NoCoverage: Incomplete and NoTiles share a retirement arm
A slow region and an ocean region are currently indistinguishable to the retirement
path. Under a large prefetch backlog, land regions with confirmed scenery coverage are
being retired to `NoCoverage` and excluded from prefetch for the remainder of the
session.

Reported by @mmaechtel from v0.4.7-alpha.1 testing —
https://github.com/samsoir/xearthlayer/issues/176#issuecomment-5307202196

## Observed

Flight LOWS → EDDF (A20N), cold cache, ~70 min. Two Baden-Württemberg regions —
47°N/8°E and 48°N/8°E, both unambiguously land — were retired as `NoCoverage` and then
disproved by on-demand generation within three minutes.

```
10:13:59 INFO Stale InProgress region demoted for retry lat=48 lon=8 attempt=1 max=3
10:16:16 INFO Stale InProgress region demoted for retry lat=48 lon=8 attempt=2 max=3
10:18:31 WARN Region marked NoCoverage after 3 failed attempts lat=48 lon=8 attempts=3
10:21:14 WARN Prefetch state contradicted by on-demand generation — region demoted
              lat=47 lon=8 tile_row=2863 tile_col=4300 tile_zoom=13 was_no_coverage=true
```

Immediately prior, prefetch had queued a backlog several times larger than one cycle
could submit:

```
10:14:16 INFO Prefetch cycle summary strategy="sliding_box" tiles_planned=6774 tiles_submitted=1001 tiles_pending=5773
```

Knock-on effect in the same window: 27 × `ERROR TIMEOUT: DDS generation exceeded 30s`
across 15 distinct tiles, **all inside the two retired regions** (col 4300 @ ZL13, col
8601 @ ZL14), each returning a magenta placeholder. The same eight tiles failed on an
exact 30-second cycle as X-Plane re-requested them.

The `PrefetchStateObserver` caught both regions and the rescue path repaired them
(`regions_nocoverage` → 0, `promotions_rescue` 0 → 2), so the safety net functioned. But
both regions were excluded from prefetch for ~3 minutes while the aircraft flew into
them.

## Root cause

`AdaptivePrefetchCoordinator::evaluate_stale_regions` merges two `RegionDiskState`
variants that mean opposite things —
`xearthlayer/src/prefetch/adaptive/coordinator/core.rs:1148`:

```rust
RegionDiskState::Incomplete | RegionDiskState::NoTiles => {
    let attempts = self.region_attempts.entry(region).or_insert(0);
    *attempts += 1;
    if *attempts >= MAX_REGION_ATTEMPTS {
        geo_index.insert::<PrefetchedRegion>(region, PrefetchedRegion::no_coverage());
    ...
```

Per `boundary_strategy.rs:34-43`:

| State | Meaning | Correct terminal outcome |
|---|---|---|
| `NoTiles` | The scenery index attributes **zero** tiles to the region | `NoCoverage` — genuinely no coverage |
| `Incomplete` | The index attributes N tiles, **at least one not yet on disk** | Retry — this is positive evidence that coverage *exists* |

The comment justifying the arm reads *"No tiles indexed: the region can never be
confirmed, so retiring it after repeated attempts is correct."* That is true of
`NoTiles` and only `NoTiles`. It was written for one state and applied to an arm holding
two.

For 47°N/8°E the scenery index was reporting *"this region has tiles"* on all three
strikes, and the region was retired as having no coverage regardless.

Note that `RegionDiskState` was introduced in #223 specifically to stop "cannot
determine" collapsing into "confirmed absent" — its doc comment says *"Deliberately not
a `bool`"*. This consumer merges two of the four variants back together. (The sibling
merge at `boundary_strategy.rs:177` is harmless: that path only ever promotes.)

## Why it fires under load, and why that is the worst time

The strike counter is a wall-clock deadline — `stale_region_timeout` (default **120s**)
× `MAX_REGION_ATTEMPTS` (3) ≈ 6 minutes — measured against a queue whose depth prefetch
itself sets. `region_disk_state` requires `.all()` tiles present, so a region 95%
downloaded is still `Incomplete` and still takes the strike. Any region entering
prefetch during a multi-thousand-tile cold-cache backlog is on a ~6-minute fuse.

Retirement is then self-reinforcing. `PrefetchedRegion::should_prefetch` is
`!index.contains::<PrefetchedRegion>(region)` (`geo_index/layers.rs:142`), so a
`NoCoverage` region is dropped from planning for the rest of the session. Prefetch stops
covering the ground the aircraft is flying into → FUSE picks the work up on demand
against an already-saturated executor → more load → more retirements.

The observer is reactive by construction: it can only fire once X-Plane has demanded a
tile in the region. An exposure window is therefore the design floor, not bad luck.

The reporter notes the host was also swapping during this window (`vm_mb` 16.9 → ~30 GB,
tracked separately). That may well be the cause of the slowness, but it does not change
the verdict — slowness is the *only* input to a decision about whether scenery exists.

## Fix direction

`NoCoverage` should be reachable only from `NoTiles`. That is what the state name
asserts and the only case where the evidence supports it.

That leaves one genuine design question, which the current merged arm was implicitly
answering: **what is the terminal state for a region that has indexed tiles but never
completes?** Retrying forever risks unbounded churn against permanently-404 tiles —
which is why the `else` branch carries its "do NOT clear `region_attempts`" comment.
Candidates:

1. Back off the stale timeout per attempt rather than striking out.
2. Strike only on *no progress since the last check* (on-disk tile count static),
   separating "slow" from "stuck".
3. Add a distinct terminal state that stops retrying without claiming absence.

To be settled before implementation.

## Acceptance criteria

1. A region whose scenery index reports ≥1 attributed tile is never marked `NoCoverage`,
   regardless of how long its tiles take to arrive.
2. A region whose scenery index attributes zero tiles still reaches `NoCoverage`.
3. Regression test: a stale `InProgress` region reporting `Incomplete` across
   `MAX_REGION_ATTEMPTS` evaluations does not end in `NoCoverage`. This test must fail
   against the current implementation.
4. A cold-cache flight with a multi-thousand-tile backlog produces zero
   `Region marked NoCoverage after N failed attempts` warnings over land, and zero
   `was_no_coverage=true` observer demotions.

## Known-issues note

The v0.4.7-alpha.1 release notes currently tell testers that `regions_nocoverage` over
land is worth reporting. Until this is fixed that will generate duplicate reports of a
known cause; worth a note on the release.

samsoir:
## Summary

A region that was merely **slow** was indistinguishable from a region with **no scenery
coverage**, so under a large prefetch backlog land regions were permanently retired as ocean and
excluded from prefetch for the rest of the session.

`AdaptivePrefetchCoordinator::evaluate_stale_regions` merged two `RegionDiskState` variants that
mean opposite things:

```rust
RegionDiskState::Incomplete | RegionDiskState::NoTiles => { /* strike -> NoCoverage */ }
```

- `NoTiles` — the scenery index attributes **zero** tiles. `NoCoverage` is correct.
- `Incomplete` — the index attributes N tiles, ≥1 not yet on disk. That is *positive evidence
  coverage exists*; `NoCoverage` is provably wrong.

The comment justifying the arm reads *"No tiles indexed: the region can never be confirmed"* —
true of `NoTiles` and only `NoTiles`. Written for one state, applied to an arm holding two.

Field evidence from @mmaechtel on v0.4.7-alpha.1 (LOWS→EDDF, cold cache): 47°N/8°E and 48°N/8°E —
Baden-Württemberg, not water — retired after 3 × 120s strikes at 10:18:31, then disproved by
on-demand generation at 10:21:14. **2m43s.** Inside those two regions: 27 × `TIMEOUT: DDS
generation exceeded 30s`, each returning a magenta placeholder.

Retirement was self-reinforcing: `NoCoverage` drops a region from planning, so prefetch stops
covering ground the aircraft is entering, FUSE picks the work up on demand against an already
saturated executor, and the added load produces more retirements.

## Changes

**The fix**

- `NoCoverage` is now reachable **only** from `NoTiles`, and immediately rather than after three
  strikes — the scenery index is fully built before the coordinator receives it, so a second
  opinion would be identical. `Incomplete` reaches `NoCoverage` by no path.
- **Strikes are progress-based.** A strike is recorded only when the region's covered-tile count
  did not increase since the previous evaluation. Slow no longer looks like absent, which is the
  reporter's own diagnosis made mechanical.
- `RegionDiskState::Incomplete` carries `{ covered, total }`. The old code reduced "387 of 400
  tiles present" to one bit, then to a strike, then to a permanent verdict.
- New **non-terminal** `RegionState::Deferred { retry_after }` with a 20s/30s/40s/60s ladder,
  60s repeating thereafter. The cap is set by the field evidence above, not by taste: the sim
  demanded that region 2m43s after prefetch gave up, so the window in which a region is skipped
  with nothing in flight has to stay well inside that.
- `should_prefetch` matches on state instead of being a bare `!contains(...)`, under which
  `InProgress`, `Prefetched` and `NoCoverage` were indistinguishable to the planner.
- **Demand beats backoff:** an on-demand FUSE generation inside a `Deferred` region clears the
  deferral immediately and does not count as a divergence. The strike count deliberately survives
  this, so the ladder keeps escalating across repeated demand-clears.
- Retry bookkeeping is pruned when a region is evicted from the retained set.

**Why `Incomplete` is never terminal**

We intend to explore falling back between imagery providers when tiles cannot be completed. That
constrains this fix: whatever state a never-completing region lands in must leave its tiles
retryable. This eliminated two otherwise-reasonable designs rather than merely ranking below them
— treating permanently-404 tiles as "covered" (region reaches `Prefetched`, which nothing
revisits) and a permanent `Partial` state both foreclose retry. `Deferred` is the state a
provider-fallback feature will attach to.

**Performance**

- `Cache::contains_sync` added; `tile_exists_blocking` no longer pays `block_in_place` +
  `block_on` + a `Box::pin` per tile. Resolves the `TODO(#175)` in `dds_disk_bridge.rs`.
- `CoverageDetail::{FirstMissOnly, ExactCounts}` keeps the every-2-second promotion path
  short-circuiting while only the rescue path counts exactly. Without it the whole-branch review
  measured ~44 filesystem `stat()` calls per absent tile landing on a hot path — the cost analysis
  in the design costed only one of the two sources `tile_is_covered` consults.

**Observability** — this is what the acceptance flight is judged on

- `Region making no progress — deferred` WARN carrying `covered`/`total`. `covered=387 total=400`
  reads as "13 tiles are 404ing"; `covered=0 total=400` reads as "nothing is arriving at all".
  Two very different faults, previously indistinguishable.
- `regions_deferred` (counter), `regions_deferred_active` (gauge) and `deferrals_cleared`
  (counter) in the periodic `Prefetch sample` line.
- `mark_no_coverage` raised to `info!` — it is now the only path that marks `NoCoverage` in
  practice, and was invisible at default log level.
- `regions_nocoverage` becomes trustworthy: it counts only regions the index says are genuinely
  empty.

## Related Issues

Fixes #226 (the keyword is also in commit `e4ace5d`, since this PR targets `develop/0.4.7`
rather than the default branch and a PR-body keyword is inert here).

Found during this work, filed separately, **not fixed here**:

- #228 — a pre-existing second path marks `NoCoverage` when there is no scenery index at all,
  bypassing the predicate entirely. Same "cannot tell ≠ absent" family as this issue. Confirmed
  reachable; can corrupt criterion F3 for a tester with no ortho packages installed.
- #229 — `AggregatedState::reset()` has no production caller, so every counter is cumulative
  since start rather than per-interval.
- #230 — the rescue path's exact-count scan can burst ~500k `stat()` calls synchronously on
  cold-cache terrain entry. Analysis and watch signature recorded; **watch for this during the
  flight** rather than treating it as backpressure.

Context: #223 introduced `RegionDiskState` specifically to stop "cannot determine" collapsing
into "confirmed absent" — its doc comment says *"Deliberately not a `bool`"*. This consumer had
merged two of the four variants back together.

## Test Plan

- [x] `make pre-commit` passes (fmt + clippy + tests) — 2499 lib tests, 63 doctests, 0 failed
- [x] `cargo test -p xearthlayer --features debug-map` — 16/16 (that feature's tests had **never
      compiled**; `RegionState` did not derive `Debug`, fixed here)
- [x] Four regression tests were verified to **fail against the pre-fix code**, two on assertions
      rather than compile errors. The load-bearing one reproduces the defect at evaluation 3 —
      the documented strike count.
- [x] Mutation-verified: removing the short-circuit, deleting either end of the metric chains, or
      reverting the `FirstMissOnly` call site each fails a specific test.

**Acceptance is a flight, not CI.** Green tests do not close #226. The corrected criteria are in
https://github.com/samsoir/xearthlayer/issues/226#issuecomment-5310644169 — note the criteria in
the issue description were **unmeasurable**, greping for a log line this PR deletes. In short,
on a cold-cache flight over terrain with packages installed:

| | Check |
|---|---|
| **F1** (decisive) | `grep -c 'was_no_coverage=true'` → **0** |
| **F2** | `Region making no progress — deferred` should **appear**; read `covered`/`total` |
| **F3** | `regions_nocoverage` flat over covered terrain |
| **F4** | `regions_deferred` is cumulative — **difference successive samples**; `regions_deferred_active` should fall back toward 0 in cruise |
| **F5** | `promotions_normal` dominates `promotions_rescue`; `state_diverged` ≈ 0 |

Copy `~/.xearthlayer/xearthlayer.log` before relaunching — it is overwritten every start.

## Checklist

- [x] Documentation updated if applicable (CLAUDE.md, docs/dev/adaptive-prefetch-design.md,
      docs/dev/geo-index-design.md)

## Known minors, accepted

- Retry entries can leak via two paths eviction cannot see (the advancing arm and the observer's
  demand-clear). ~50 bytes each, low thousands worst case; self-correcting after one strike now
  that `last_covered` tracks the latest observation rather than a high-water mark.
- `regions_deferred_active` also counts regions whose deferral has expired but which have not yet
  been re-marked `InProgress`, so under sustained backpressure it reads slightly high.

No `[Unreleased]` CHANGELOG entry — by convention here the changelog is compiled at release-cut
time, so an empty section is correct in a feature PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -177,7 +177,8 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
     - `GeoLayer` trait - Marker trait for storable layer types (`Clone + Send + Sync + 'static`)
     - `PatchCoverage` - Layer type indicating a region is owned by a scenery patch
     - `RetainedRegion` - Layer type tracking regions in the SceneryWindow's retained set
-    - `PrefetchedRegion` - Layer type tracking prefetch state per region (InProgress, Prefetched, NoCoverage)
+    - `PrefetchedRegion` - Layer type tracking prefetch state per region (InProgress, Prefetched, NoCoverage, Deferred)
+    - **Region retirement (#226)**: `NoCoverage` is reachable *only* when the scenery index attributes zero tiles to a region. A region with indexed tiles that is merely slow or stuck is `Deferred` — a non-terminal state that expires on a 20/30/40/60s ladder — so its tiles stay retryable
     - Locking: `RwLock<HashMap<TypeId, DashMap>>` — rare write locks, concurrent reads via DashMap shards
     - `populate()` provides atomic bulk loading (builds outside lock, swaps in)
     - Used by FUSE (`is_geo_filtered`), prewarm, and prefetch for patch region exclusion and prefetch state tracking
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,7 +126,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
     - `PrefetchBox` - Heading-biased sliding box (speed-proportional extent, proportional 80/20 bias, cruise phase)
     - `ExtentCalculator` - Speed-proportional box extent: clamped linear ramp 3.5° at 40kt to 6.5° at 450kt
     - `SceneryWindow` - Window config and dimensions holder
-    - `BoundaryStrategy` - Region lifecycle management (InProgress/Prefetched/NoCoverage)
+    - `BoundaryStrategy` - Region lifecycle management (InProgress/Prefetched/NoCoverage/Deferred)
     - `PhaseDetector` - Ground/Cruise flight phase state machine; reset via `on_ground` from AircraftState on telemetry resume
     - `PerformanceCalibrator` - Measures throughput during initial load for mode selection
     - `SimState` - Direct sim state detection via X-Plane Web API (paused, on_ground, scenery_loading, replay)
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -256,7 +256,7 @@ This follows a **two-phase commit** pattern:
    | `RegionDiskState` | Outcome |
    |--------------------|---------|
    | `Complete` | Promote straight to `Prefetched` (the rescue path; see [Region Tile-Set SSOT](#region-tile-set-ssot) below); clear the region's strike count |
-   | `Incomplete { covered, total }` | **Progress-based, not time-based.** If `covered` grew since the previous evaluation, the region is advancing: clear strikes, revert to *absent* and retry immediately. If `covered` did not grow, it's a strike: increment the strike count and mark `Deferred { retry_after: now + deferral_for(strikes) }` |
+   | `Incomplete { coverage }` | Exact `covered`/`total` counts on this path — it asks for `CoverageDetail::ExactCounts`. **Progress-based, not time-based.** If `covered` grew since the previous evaluation, the region is advancing: clear strikes, revert to *absent* and retry immediately. If `covered` did not grow, it's a strike: increment the strike count and mark `Deferred { retry_after: now + deferral_for(strikes) }` |
    | `NoTiles` | The *only* path to `NoCoverage` — the scenery index attributes zero tiles to this region, which is definitive on first look (the index is fully built before the coordinator sees it). Retire to `NoCoverage`, clear the region's strike count |
    | `Unknown` | No authoritative source to consult (no `DdsDiskCacheChecker` and/or no `SceneryIndex`) → left `InProgress` untouched; this does **not** count as a strike, since "cannot tell" must not be treated the same as "checked and it is absent" |
 
@@ -321,6 +321,27 @@ answer feeds a decision whose terminal outcome is permanent exclusion
 no `SceneryIndex` wired) must stay distinguishable from "we checked and it is
 absent" (`Incomplete`/`NoTiles`), or a region could be retired to `NoCoverage`
 on a cycle where it was never actually checked.
+
+**How thoroughly it scans is a parameter, not a second function.** Callers pass
+a `CoverageDetail`. The rescue path asks for `ExactCounts` because `covered` is
+what separates an advancing region from a stuck one, and it can afford the full
+scan: it only ever looks at regions already stale for `stale_region_timeout`
+(120s). `promote_completed_regions` asks for `FirstMissOnly` and stops at the
+first absent tile, because it discards the counts and runs on every coordinator
+cycle (2s) over every `InProgress` region. The difference is not academic: when
+the XEL DDS cache misses, the disjunction below falls through to
+`OrthoUnionIndex::dds_tile_exists`, which tries four filename prefixes and —
+because `textures/` is in `LAZY_DIRECTORIES` and is never indexed — resolves
+each by `stat()`ing every installed source, so one absent tile costs roughly
+4 × N_sources syscalls. Counting a cold region's whole tile set every 2s steals
+CPU from the executor in exactly the backlog condition this machinery exists to
+relieve.
+
+A `FirstMissOnly` answer reports `Incomplete { coverage: TileCoverage::NotCounted }`.
+The variant carries no payload, so "not counted" cannot be mistaken for the
+observation `covered: 0` — a fabricated zero would read as "nothing has
+arrived" and drive a strike. `TileCoverage::counts()` returns an `Option`,
+forcing every reader to handle the absence explicitly.
 
 **Coverage is a disjunction.** A tile counts as covered if it is present in
 the XEL DDS disk cache (`DdsDiskCacheChecker`) **or** ships inside an
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -236,14 +236,15 @@ If non-empty → expand to DDS tiles, filter, submit
 
 ### Region States (GeoIndex PrefetchedRegion layer)
 
-Each DSF region in the XEL Window has one of four states:
+Each DSF region in the XEL Window has one of five states:
 
 | State | Meaning | Prefetch Behaviour |
 |-------|---------|-------------------|
 | *absent* | Never evaluated | Include in target diff → submit |
 | `InProgress` | Tiles submitted, awaiting completion | Skip bulk submission (prevents duplicate work) |
 | `Prefetched` | All tiles confirmed cached | Exclude from target diff entirely |
-| `NoCoverage` | No scenery data in this DSF region | Exclude from target diff, skip silently |
+| `NoCoverage` | Scenery index attributes zero tiles to this region | Exclude from target diff, skip silently |
+| `Deferred { retry_after }` | Has indexed tiles but made no progress; non-terminal | Excluded from target diff until `retry_after`, then eligible again — the one state that *expires* |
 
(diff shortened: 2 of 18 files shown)


samsoir:
## Acceptance flight 1 — passed

Flown on this branch. KDEN eastbound, ~2h. Terrain flown was **not** previously cached:
**6,832,128 chunks downloaded** (26,688 tiles, 254 GB DDS written), opening backlog **8040 tiles
planned / 7039 pending** — larger than the backlog in the original report.

| | Check | Result |
|---|---|---|
| **F1** | `was_no_coverage=true` | **0** ✅ |
| **F2** | deferral path exercised | **0 — never fired** ⚠️ |
| **F3** | `regions_nocoverage` | **0** across all 120 samples, and 0 `marked NoCoverage` lines ✅ |
| **F4** | deferrals bounded | 0 / 0 — trivially, unexercised |
| **F5** | `promotions_normal` : `promotions_rescue` = **141 : 1**; `state_diverged=0`; `regions_demoted=0` | ✅ |

Also **zero `ERROR` lines and zero `TIMEOUT`s** — the original report had 27 — and
`fuse_generations` flat at 1993 for 100 minutes while prefetch stayed active over the same
window, so prefetch was ahead of the aircraft rather than the sim being paused.

**F2** is a property of the design rather than a gap in the test: on a region's first stale
evaluation `last_covered` starts at 0, so any coverage at all routes to the advancing branch.
Reaching a deferral needs two consecutive stale windows with zero net progress. Every region here
was gaining tiles. The originally-reported regions still defer under this rule — they sat at
`covered=0`.

Full write-up: https://github.com/samsoir/xearthlayer/issues/226#issuecomment-5336100882

**The backoff ladder itself remains unvalidated in flight** — exercising it needs starvation, not
merely a cold cache.

PR #238, opened 2026-08-22
fix(release): template the AUR PKGBUILD from source and build it in CI (#222)

### LINKED ISSUE #222: [Bug]: Creating Error AUR Build - CashyOS
### Bug Description

I can not compile XearthLayer. I attach the build termin log.

[error.txt](https://github.com/user-attachments/files/30947517/error.txt)

### Steps to Reproduce

Try to build with CashyOS

### Expected Behavior

no

### Actual Behavior

no

### System Diagnostics

```text
no avaiable
```

### Relevant Logs

```text

```

### Screenshots

_No response_

### Severity

Crash / Data Loss

### Checklist

samsoir:
Fixes #222

`release.yml` re-authored the PKGBUILD from a heredoc instead of templating
`pkg/arch/PKGBUILD`, and that second copy never carried `options=(!lto)`. On
Arch and CachyOS the `cc` crate picks up `-flto=auto` from `CFLAGS`, ring's C
sources compile to GCC slim-LTO objects with no machine code, and `lld` cannot
read GCC GIMPLE — so every ring symbol comes back undefined. The release asset
is the only route to this package, so every Arch-family user hit it.

## Changes

- **PKGBUILD is templated by sed** from `pkg/arch/PKGBUILD`, as the Makefile and
  the RPM job already do. The duplicate is gone, so this drift cannot recur.
- **`prepare-aur` runs in `archlinux:base-devel`**, so `.SRCINFO` is generated by
  `makepkg --printsrcinfo` rather than being a third hand-written copy.
- **`makepkg -Cf -s` runs as a release-blocking gate.** Asserting the flag is
  present does not prove the package links; building it does.
- `check()` dropped from the packaged PKGBUILD — a package build is not a test
  run, and two tests are known to be sensitive to CPU count and RAM.
- `make aur-prepare` repaired: `$(eval $(shell sha256sum ...))` expanded before
  `curl` ran, so it checksummed a file that did not exist yet. It has been
  broken; releases went through the duplicate path, so nobody noticed.

## Verification

Mutation-tested on CachyOS with a stock `makepkg.conf` (LTO on), v0.4.6 source:

| | result |
|---|---|
| with `options=(!lto)` | exit 0 — `Finished making: xearthlayer 0.4.6-1` |
| without | exit 4 — `rust-lld: undefined symbol: ring_core_0_17_14__*`, 21 errors |

The failing case reproduces the reporter's log exactly, so the CI gate
discriminates rather than merely passing.

To repeat it: `make aur-prepare TAG=v0.4.6 && cd pkg/aur && makepkg -Cf`

Note that **no preview tag can exercise this job** — `prepare-aur` is skipped
for pre-releases — so the local run above is the only pre-merge proof available.

## Two things worth a reviewer's opinion

1. The build gate is **release-blocking** (`publish` requires
   `prepare-aur != failure`). Deliberate: shipping an unbuildable PKGBUILD is
   the bug this job exists to prevent. The cost is that a transient
   pacman-mirror failure stops a release; recover with
   `gh run rerun <id> --failed`, since the tag is immutable.
2. `pkg/arch/PKGBUILD-git` still has `check()`. Dormant and outside the release
   path, so left alone rather than changed silently.

`[Unreleased]` is intentionally untouched — the CHANGELOG is compiled at
release-cut time.


pkg/arch/PKGBUILD
```diff
@@ -1,6 +1,6 @@
 # Maintainer: Sam de Freyssinet <sam@def.reyssi.net>
 pkgname=xearthlayer
-pkgver=0.2.0
+pkgver=0.4.6
```
samsoir: This should be pinned to Cargo.toml `workspace.package.version`

samsoir (commented):
One note to fix

pkg/arch/PKGBUILD
samsoir: Fixed in `e2be420`. You were right that this was the same defect wearing a different hat — I swapped a stale `0.2.0` for a hand-written `0.4.6`, which is still a second place to remember.

Rather than set the number, I asserted it: `xearthlayer/tests/packaging_metadata.rs` requires `pkgver` to equal `workspace.package.version` with the prerelease suffix stripped (Arch forbids the hyphen), so `0.4.7-alpha.3` packages as `0.4.7`. It now runs under `make pre-commit`, so a Cargo.toml bump that forgets the PKGBUILD fails locally.

Two more invariants went in the same file while it was cheap — no hyphen in `pkgver`, and `options=(!lto)` present. The second is the #222 regression itself, now caught by the test suite rather than by a user's failed build. Both mutation-checked: the pkgver test fails on `0.4.6`, the lto test fails when the line is deleted.

One consequence worth naming: `pkgver` now points at a version that may not be tagged yet, so a direct `makepkg` against the canonical file 404s until release. I think that is the safer failure — the old stale value pointed at a *real* but ancient tag and would have silently packaged the wrong source. Release tooling seds `pkgver` from the tag, so shipped artifacts are unaffected.

Issue #227, opened 2026-08-16
Unbounded memory growth proportional to tiles processed — OOM kill on long flights


samsoir:
Memory commitment grows monotonically and without bound in proportion to the number of
tiles processed. On a 12-hour flight this reached 64 GB anon and the kernel OOM killer
terminated the process. The growth is reproducible on three flights, two machines and
two providers.

Successor to #209, which shipped the telemetry (PR #215) needed to diagnose this but
deliberately did not attempt a fix. This issue is the diagnosis and the fix.

## Symptom

2026-07-27, v0.4.6, 12h long-haul (KZ → Texas). Killed by the OOM killer at 16:54:55Z —
the last log line in the run. No application error; nothing in the app log indicated a
problem.

From `journalctl -k`, not the app log:

- xearthlayer: **64.0 GB anon** (9.3 GB `rss_anon` + 54.7 GB swapped), 70.8 GB VM
- Host: 91 GiB RAM, swap is **zram only** (RAM-backed) — 176 kB free at kill
- 120,999 jobs submitted == started == completed — **no job backlog**
- 25,647,104 chunk downloads

85% of the memory was swapped out: written once, never re-read.

## The growth is a function of work done, not elapsed time

This is the central finding, and it is what the telemetry from PR #215 bought us. Three
independent runs, normalised two ways:

| Run | Duration | Mchunks | VM rise | **GB/hour** | **GB/Mchunk** | MB/tile |
|---|---:|---:|---:|---:|---:|---:|
| LOWS→EDDF (@mmaechtel, cold cache) | 73 min | 8.29 | 29.9 GB | **24.5** | **3.60** | 0.94 |
| July OOM run (at kill) | 719 min | 25.65 | 60.5 GB | **5.1** | **2.36** | 0.62 |
| KDEN→YPPH (uptime 1500–21000) | 325 min | 5.40 | 11.5 GB | **2.1** | **2.14** | 0.56 |

Per hour, the rate varies **12×**. Per million chunks processed, it varies **1.7×**.
Elapsed time does not predict the growth; throughput does.

@mmaechtel's run contains the control that makes this conclusive — two idle rows while
the aircraft sat on the ground with nothing downloading or encoding:

```
uptime  rss_mb  vm_mb  chunks_ok
 1200s   12946  17858  2,827,116
 1800s   12968  16844  2,855,680   <- idle on ground
 2400s   13000  16908  2,855,680   <- idle on ground
[...]

PR #239, opened 2026-08-22
refactor(metrics): delete AggregatedState::reset(), which was never called (#229)

### LINKED ISSUE #229: AggregatedState::reset() has no production caller — counters are cumulative but documented as per-interval

samsoir:
Fixes #229

`AggregatedState::reset()` had no production caller — all three call sites were
inside `#[cfg(test)]` — so every field it zeroed has always been cumulative
since process start, while the code, its tests and several comments read as
though a per-interval reset ran.

## Why delete rather than wire it up

The issue framed this as a choice. Investigation made it one-sided: **`reset()`
zeroes eight in-flight gauges alongside the counters.**

| Field | Maintained by |
|---|---|
| `downloads_active`, `disk_writes_active`, `mem_cache_writes_active` | `+= 1` / `saturating_sub(1)` |
| `jobs_active`, `encodes_active` | `+= 1` / `saturating_sub(1)` |
| `fuse_requests_active`, `fuse_requests_waiting` | `+= 1` / `saturating_sub(1)` |
| `memory_cache_size_bytes` | assigned wholesale |

Zeroing an increment/decrement gauge mid-flight strands the pending decrements
at the `saturating_sub` floor, so the gauge under-reports by the in-flight depth
— every interval, cumulatively. Calling `reset()` would have taken eight working
gauges and broken them; "wire it up" was really "audit 45 fields and split
gauges from counters first".

Cumulative is also the better contract: a cumulative series can be differenced,
a windowed one cannot be un-differenced, and a dropped sample loses that
interval permanently. Keeping it also means existing flight logs stay comparable
with new ones, which matters while #227 is being diagnosed from four captured
runs.

## Coverage

`test_aggregated_state_reset` seeded only three genuine counters and never
touched a gauge, so it could not have detected the defect above. It and
`test_promotion_counters_reset` are removed as tests of dead code.

The production gauge-vs-counter distinction from #226 keeps full coverage:
counters accumulate (`test_promotion_counters_accumulate`), gauges are assigned
wholesale (the `PrefetchRegionState` replace-not-accumulate test at
`daemon.rs:1458`). Only assertions about the dead function's behaviour were
removed.

## Verification

`make pre-commit` green — 2,494 tests, 93 in `metrics`. −116 / +34 lines.

Acceptance criteria from the issue: `reset()` is out of the tree; no comment
describes a field as per-interval; `docs/dev/memory-telemetry.md` and
`docs/dev/adaptive-prefetch-design.md` now state that counters are cumulative,
gauges are current-state, peaks are high-water marks, and that a rate comes from
differencing successive samples.

## Note for reviewers

No behaviour changes — this function never ran in production. What changes is
the documented contract, which now matches what the counters have always done.

`[Unreleased]` is intentionally untouched; the CHANGELOG is compiled at
release-cut time.

### DIFF
--- a/docs/dev/adaptive-prefetch-design.md
+++ b/docs/dev/adaptive-prefetch-design.md
@@ -444,8 +444,9 @@ surface.
 
 The daemon aggregates these and logs them on the same 60s cadence as the
 memory sample. The counters among them are cumulative totals since process
-start, because `AggregatedState::reset()` — which does zero them — has no
-production caller; wire it up and these become totals since the last reset:
+start; nothing windows them, so difference successive samples to get a rate.
+The `regions_*` state fields are gauges, assigned wholesale from the GeoIndex
+each maintenance cycle rather than accumulated:
 
 ```
 INFO Prefetch sample uptime_s=... regions_in_progress=... regions_prefetched=...
--- a/docs/dev/memory-telemetry.md
+++ b/docs/dev/memory-telemetry.md
@@ -82,6 +82,27 @@ mebibyte", not necessarily "nothing".
 | `dds_handles_peak` | `state.fuse_handles_peak_open` | Highest concurrent open count this session. **Read this, not `dds_handles_open`, when sizing the cap.** The current gauges are sampled on open, on tile production and on release, so a 60-second reader almost always catches them just after a release: the first KDEN run reported `dds_pinned_mb=0` throughout while 31 files were open. |
 | `dds_pinned_peak_mb` | `state.fuse_handles_peak_pinned_bytes` | Highest pinned total this session. Compare against `MAX_PINNED_TILE_BYTES`: if it approaches 512, `open()` is close to silently dropping memoisation and the cap wants raising. |
 
+### Counters are cumulative; gauges are current-state
+
+Every counter on this line — `tiles_done`, `chunks_ok`, `chunks_failed`,
+`gc_evicted_mb`, `fuse_*_reads`, and the `Prefetch sample` counters
+(`regions_deferred`, `deferrals_cleared`, `promotions_*`) — is **cumulative
+since process start**. Nothing windows them: there is no per-interval reset, so
+a raw value always climbs and a single line cannot tell you what happened in the
+last minute. **Difference successive samples to get a rate**, or divide by
+`uptime_s` for a session average.
+
+Gauges — `encodes_active`, `disk_writes_active`, `mem_cache_writes_active`,
+`dds_handles_open`, `dds_pinned_mb`, `chunk_index_entries`, the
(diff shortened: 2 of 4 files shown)


samsoir:
:+1:

Issue #234, opened 2026-08-19
Virtual DDS reads clone the whole tile once per ≤1 MiB range request


samsoir:
## Problem

Virtual DDS files are opened with `FOPEN_DIRECT_IO` (#65), so the kernel serves nothing from
its page cache and every X-Plane read reaches our handler. That is deliberate and correct —
it is what keeps `FuseLoadMonitor`, `SceneTracker` and `DdsAccessEvent` seeing the real access
pattern. The problem is what the handler does with each of those calls: `request_dds()` returns
an owned `Vec<u8>` containing the **whole tile**, and the handler copies out the ≤1 MiB range
the kernel asked for and drops the rest.

Because the kernel caps every FUSE read at `max_pages × PAGE_SIZE` (1 MiB), a single tile is
never delivered in one call. Measured with a probe FS on our own `fuse3` fork, an 11.17 MB file
takes **12 read calls** when the reader issues one whole-file `read(2)`, and **23** for an
ordinary sequential reader. Serving one texture therefore allocates, fills and drops an
11.17 MB buffer 12–23 times — 134–257 MB of copying to deliver 11.17 MB — even on a pure
memory-cache hit, where the bytes are already resident and only need to be borrowed.

This is the hottest path in the program: it runs for every tile X-Plane draws. The repeated
11.17 MB allocations sit just under glibc's 32 MB adaptive `mmap` threshold, which is the
mechanism behind the leading candidate in #227, and they are bounded by the FUSE disk I/O
limiter (up to 256 permits) rather than by `max_concurrent_jobs`.

## Sites

- `xearthlayer/src/fuse/fuse3/ortho_union_fs.rs:743` — read handler for virtual DDS inodes
- `xearthlayer/src/fuse/fuse3/shared.rs:345` — `request_dds_impl` returns `Vec<u8>` by value
- the memory/DDS-disk cache bridges and the request coalescer's `into_data()` sit on the same path

## Notes

`FOPEN_DIRECT_IO` should stay: this is about avoiding a copy per call, not about restoring
kernel caching, and every read continues to reach the handler and fire its telemetry.
Overlaps with #175 (`CacheLocator`), which owns the cache-access seam this path goes through.

## Related

#227 (unbounded memory growth) · #175 · #65 · #233

samsoir:
## Scope addition: FUSE read metrics

Read instrumentation is folded into this pass alongside #233, so the fix ships with the
instrument that proves it worked. The full definition is in
[...]

PR #240, opened 2026-08-22
feat(fuse): count and announce DDS handle budget exhaustion (#236)

### LINKED ISSUE #236: MAX_PINNED_TILE_BYTES is a guess, and its fallback is silent
## Problem

`MAX_PINNED_TILE_BYTES` in `Fuse3OrthoUnionFS` caps the DDS bytes that open file
handles may pin at 512 MiB. Past that ceiling `open()` stops handing out memoising
handles, and reads fall back to resolving the tile per call — the pre-#234 cost. The
degradation is silent apart from a `debug!` line, so on a scene dense enough to reach
the ceiling the fix quietly stops applying and the only visible symptom is that
`fuse_dds_alloc_mb` climbs again.

The number was chosen without data, because nothing measured how many DDS files
X-Plane keeps open at once. First measurement (KDEN scene load, 2026-08-19) put
concurrent opens at **31**, so a peak of roughly **330 MiB against a 512 MiB ceiling** —
about 1.5× headroom on a single mid-size airport. A denser scene, a larger scenery
window, or several packages loading at once could plausibly exceed it.

## What is needed

Size the cap from flight data rather than from a guess, using `dds_handles_peak` and
`dds_pinned_peak_mb` on the periodic `Memory sample` line (added with #234). Decide
whether the ceiling should be a constant at all, or derived — from the memory cache
budget, from `fuse.max_background`, or made configurable. If it stays a constant, the
fallback should be observable at default log level rather than only under `--debug`,
so a flight can tell the difference between "the fix is not working" and "the cap
engaged".

## Related

#234 (introduced the cap) · #227 (why new retention was bounded in the first place)

samsoir:
Fixes #236

Past `MAX_PINNED_TILE_BYTES`, `open()` stops handing out memoising handles and
reads fall back to resolving the tile once per call — the pre-#234 cost. That
was announced by one `debug!` line, so on an ordinary flight the fix could stop
applying with no visible symptom beyond `fuse_dds_alloc_mb` climbing again,
which reads identically to "#234 never worked".

## Changes

- **`dds_budget_exhausted`** on the `Memory sample` line: cumulative count of
  opens refused a handle. Non-zero means the cap engaged; zero while
  `fuse_dds_alloc_mb` climbs means a different fault.
- **Latched `warn!`** at default level on the first refusal. Latched because the
  cap engages for every open in a burst, and a per-open warn would bury the log
  it exists to serve.
- **Cap logged once at mount** (`dds_pinned_cap_mb`), so a log is
  self-describing and a reader need not know which build produced it.

## The cap stays a constant, and its premise needed correcting

The issue asked whether the ceiling should be derived or configurable. It stays
512 MiB: `fuse.max_background` bounds requests *in flight* while a handle
persists *between* reads, so it does not bound pinned bytes; the memory-cache
budget is an unrelated concern; and a config key is a knob nobody can set for a
value with this much headroom.

**This much headroom is the correction.** The issue reasoned from 31 concurrent
opens to "roughly 330 MiB against 512 — about 1.5× headroom". The peak gauge
added alongside it measured **10–21 MiB**.

Handles are counted at `open()`. Bytes are counted when a tile is
**materialised**, inside `get_or_init` on the first read. X-Plane reads each
texture in about two calls and releases promptly, so 31 open files held roughly
**two** resident tiles. Multiplying the open count by the tile size overstated
the requirement by ~16×; real headroom is about **24×** (the cap holds 48 tiles).

The constant's doc comment and `docs/dev/memory-telemetry.md` now both say:
size against `dds_pinned_peak_mb`, never against `dds_handles_peak`.

## Documented rather than fixed

The cap is **soft** — admission tests `pinned + expected` without reserving, so
concurrent opens can pass together and overshoot. Reserving on `open()` would be
worse: X-Plane may open a file it never reads, so reservations would be held for
tiles that never exist and the cap would engage falsely. Immaterial at the
measured headroom; the new counter is what reports if that changes.

## Verification

`make pre-commit` green — 2,499 tests. Four new tests, all mutation-verified:

| Mutation | Test that fails |
|---|---|
| latch removed (warn every time) | `test_budget_warning_is_latched`, `test_claim_budget_warning_is_one_shot` |
| metrics emit removed | `test_open_past_budget_counts_every_refusal` |
| daemon counter assigns instead of `+=` | `test_handle_budget_exhausted_accumulates` |

The budget refusal path had **no test at all** before this; `test_open_past_budget_refuses_memoisation` now covers it.

The latch test initially asserted the flag rather than the claim and passed
under the first mutation above. `claim_budget_warning()` was extracted so the
one-shot contract is the thing asserted.

## Note on the evidence

The 10–21 MiB figures come from the session record of two scene loads (KDEN
warm, KSLC cold, 2026-08-19), not from a file that still exists —
`~/.xearthlayer/xearthlayer.log` is overwritten each run. They are stated in the
docs as those two loads on one machine, not as a general result. The counter
this PR adds is partly so the next flight leaves durable evidence.

`[Unreleased]` is intentionally untouched; the CHANGELOG is compiled at
release-cut time.

### DIFF
--- a/docs/dev/memory-telemetry.md
+++ b/docs/dev/memory-telemetry.md
@@ -46,10 +46,13 @@ so that a single unattended flight discriminates between them:
 2026-08-06 06:20:33Z  INFO Memory sample uptime_s=60 rss_mb=873 vm_mb=1459 swap_mb=0 threads=71 tiles_done=0 encodes_active=0 chunks_ok=0 chunks_failed=0 mem_cache_mb=0 dds_disk_mb=158268 chunk_disk_mb=55301 gc_evicted_mb=0 chunk_index_entries=3803083 disk_writes_active=0 mem_cache_writes_active=0
 ```
 
-(Captured before #233; a current line also carries the six `fuse_*` read
-fields described below.)
+(Captured before #233. The example above shows 16 fields; a current line
+carries **27** — the six `fuse_*` read fields, the four `dds_handles_*` /
+`dds_pinned_*` fields, and `dds_budget_exhausted`, all described below. If you
+are editing this paragraph, count the fields in `log_memory_sample` rather than
+adjusting the number by hand: it has gone stale twice that way.)
 
-Twenty-two fields, in emission order. Every `_mb` value is truncating integer
+Fields are listed below in emission order. Every `_mb` value is truncating integer
 division by 1 MiB (1 048 576 bytes), so a value of `0` means "under one
 mebibyte", not necessarily "nothing".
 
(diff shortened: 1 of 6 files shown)


samsoir:
:shipit:

PR #243, opened 2026-08-25
fix(runtime): cap the tokio blocking pool from executor capacities, remove the inert disk_io_profile (#227)

samsoir:
## The fix

The tokio blocking pool carried tokio's 512-thread default. It hosts DDS
encoding and disk I/O, so under sustained prefetch it saturates — a 12 h
alpha.3 flight found 581 threads in 103 of 730 samples.

That ratchets glibc's arena. Each excursion recruits arenas, each keeps its own
top-chunk high-water mark, and retiring threads leave those peaks in place. The
arena reached 6,166 MB while holding 572 MB of live data; all 4.6 GB of growth
was free list. Thread stacks cost a further 2.21 MB each.

The pool is now capped at `ResourcePoolConfig::blocking_threads_required()` —
`cpu + disk_io + BLOCKING_POOL_HEADROOM`, 136 on a 32-core host. The CPU and
disk I/O pools both dispatch through `spawn_blocking`, so their sum is the
worst-case demand; the reserve covers the ungated `spawn_blocking` calls
(startup scans, GPU worker, fire-and-forget cache writes).

Sizing from the storage profile instead would under-provision: the SSD profile
gives 64 against 104 of pooled demand.

## Cleanup

`cache.disk_io_profile` never affected I/O concurrency. Three places computed a
profile-based limit and nothing consumed any of them: the executor pool took a
flat `DEFAULT_DISK_IO_CAPACITY`, `ResourcePoolConfig::with_disk_io_profile()`
had no production callers, and the profile-sized `StorageConcurrencyLimiter`
`ServiceBuilder` built sat behind `#[allow(dead_code)]`.

Removed, with a `DEPRECATED_KEYS` entry so existing configs are cleaned on
upgrade — along with the wizard prompt that asked users to pick a value that
did nothing, both sizing APIs, and four `StorageConcurrencyLimiter`
constructors with no references anywhere. Of seven constructors, one was used.

Storage *detection* survives and still feeds the hardware report. With
`DiskIoProfile` gone, `config/storage.rs` held only hardware detection, so it
moved to `system/storage_detect.rs` returning `StorageType` directly.

**The live FUSE limiter at `ortho_union_fs.rs:928` is unchanged** — it gates
every passthrough read and was never profile-based.

Net −315 lines.

## Verification

`make pre-commit` green. Two mutations checked, each caught only by its own
guard:

| mutation | test |
|---|---|
| delete `.max_blocking_threads(...)` | `blocking_pool_is_capped_at_the_derived_limit` (20 threads ran vs 16) |
| drop `disk_io` from the derivation | `derived_cap_covers_what_the_pools_can_grant` |

The second is the anti-drift guard: raising `DEFAULT_DISK_IO_CAPACITY` or the
CPU multiplier without revisiting the pool now fails the suite.

Also verified green single-threaded (rules out order-dependence) and under
24-way CPU contention.

**One unexplained result:** `make pre-commit` failed once with
`2489 passed; 1 failed`. The test could not be named — the panic-handler banner
prints "FAILED" for `#[should_panic]` tests and swallowed the summary — and it
has not recurred in nine subsequent full runs. Reporting rather than omitting.

## Flight test

Not merge-ready on green CI. This changes runtime behaviour under load.

Expected: threads **581 → ~168**, roughly 980 MB less in stacks, and far fewer
arenas recruited per excursion.

The prediction to test is that `heap_mb` **stops stepping**. Watch `heap_mb`
against `threads`, not committed memory against time — the plateaus between
steps ran 45 minutes and 7,000 encodes on the last trace, long enough to look
like convergence and not be. Give it more than 40,000 encodes before believing
a plateau.

Watch for a throughput regression during prefetch bursts: 136 blocking threads
against 104 of pooled demand should be sufficient, but that is arithmetic, not
a measurement.

CHANGELOG is compiled at release time, so the empty `[Unreleased]` section is
correct here.


samsoir:
## Flight test result — passes, merging recommended

FAOR→EGLL, **11 h 03 min, 121,641 encodes, 31.1 M chunks**. That is 35% more
work than the whole alpha.3 baseline flight (90,308 encodes), so the comparison
below is at matched encode counts, not matched time.

### The cap is applied

Peak threads **581 → 205** — `205 = ~69 base/workers + the 136-thread cap`.
Exactly the intended arithmetic, confirmed in flight.

### Memory at matched work

| encodes | baseline | this build | delta |
|---|---|---|---|
| 5,000 | 6,473 (heap 1,568) | 5,026 (heap 824) | −22% |
| 40,000 | 8,640 (heap 2,971) | 7,258 (heap 1,914) | −16% |
| 55,000 | 9,812 (heap 3,872) | 7,209 (heap 2,304) | −27% |
| 90,000 | 11,936 (heap 6,192) | 9,441 (heap 3,764) | **−21%** |

All MB, committed = `anon + swap`.

The arena ratchet, measured as a rate rather than a total:

| | steps | mean step | total ratchet | per 1k encodes |
|---|---|---|---|---|
| baseline | 12 | 396 MB | 4,762 MB | 52.7 |
| this build | 18 | 299 MB | 5,393 MB | **44.3** |

### What it does not do

The mechanism is slowed, not stopped. The run ended at 11,580 MB against the
baseline's 11,876 — the same ceiling, reached after 35% more work. Arena
high-water ratcheting remains the dominant growth term. This PR is an
improvement, not a fix for #227; details and the next lever are on that issue.

### One new cost, bounded

Live arena data (`heap − free`), sampled at idle only (no active encodes, no
pending writes, threads < 100) so bursts do not contaminate it:

```
             10k    21k    40k    61k    81k   101k   120k
baseline      615    648    686    679    643     —      —
this build    331    409    567    750  1,040  1,030  1,065
```

It climbs to ~1,040 MB by 81,000 encodes and then flattens. A higher plateau,
roughly +400 MB against the baseline's ~650, not unbounded growth. Consistent
with queued-work retention reaching a steady backlog depth: capping the pool
makes work queue rather than run, and a queued `spawn_blocking` closure holds
its captures while it waits. The trade is about 6:1 in our favour — −2,428 MB
of arena against +400 MB of live.

### No regressions

- 113 failed chunks out of 31.1 M, all Apple CDN `HTTP 404` on specific tiles.
  Handled correctly — served to X-Plane, not cached.
- No throughput regression observed: ~3.1 encodes/sec sustained.
- CI green.

### Telemetry gap found in passing

An observed ~2 GB transient spike during a failure burst left **no trace** in
the sample line. Every field is an instantaneous gauge read once per 60 s, so a
spike that rises and falls inside one interval is invisible. Since the OOM
killer scores the peak rather than the average, a high-water `anon_peak_mb`
field is worth adding. Filing separately; not a blocker here.

PR #244, opened 2026-08-26
fix(runtime): cap glibc arenas at 4, with environment override (#227)

samsoir:
## The fix

glibc's arena ceiling defaults to `8 x ncores` — 256 on a 32-core host. A thread
keeps whichever arena it first got, each arena grows to cover the worst burst
*it* ever saw, and glibc returns memory only by trimming the **top** of a heap,
so space stranded below a live allocation stays committed for the life of the
process.

Total arena is therefore the **sum of every arena's personal high-water mark**.
That is why it never converged: with 256 available, every burst can recruit
*fresh* arenas that have not yet seen one, so there is always a new peak to add.

`mallopt(M_ARENA_MAX, 4)` removes the supply. Once each of four arenas has seen
its worst burst, no fresh ones remain and the sum saturates.

## In flight now — 7 h 40 min, 88,258 encodes

Same route as #243's verification flight (FAOR→EGLL), so the comparison holds
everything but this change constant.

**The arena is bounded.** It reached 3,916 MB in the first six minutes, in six
steps, and has not moved since:

```
heap since last step: min=3915  max=3916  DRIFT=1 MB
                      over 454 samples / 81,121 encodes / 7.6 h
```

It did not merely slow — it went *down* by a megabyte. Uncapped on the same
route: 18 steps spread across 11 hours, still stepping at hour 10.75, ending at
6,091 MB while holding 1,002 MB of live data.

3,916 MB is peak concurrent demand (3,093 MB measured on the baseline flight)
plus about 27% fragmentation headroom. The arena is now sized by what the
workload needs at once rather than by its history — which makes it a number that
can be stated in advance, and therefore something a `MemoryHigh=` could be set
against in v0.5.0.

### Not yet ahead on total memory

| encodes | no cap | ARENA_MAX=4 | gap |
|---:|---:|---:|---:|
| 40,000 | 7,258 | 8,834 | +1,576 |
| 55,000 | 7,209 | 8,835 | +1,626 |
| 70,000 | 9,012 | 9,613 | +601 |
| 85,000 | 9,024 | 9,716 | +692 |

The cap front-loads the whole arena into the first six minutes, so it is behind
early and the gap closes as the uncapped run accumulates. Growth rates have
separated: **19.6 MB per 1,000 encodes capped, against 71 MB per 1,000** over
the uncapped run's own tail. Where the lines cross will be in the final metrics;
it has not happened yet.

Component decomposition at idle shows why the remaining growth is not the arena:

```
enc      committed   arena   memcache   mmap-cache   live
15,957      8,653     3,916     4,085         653     399
45,594      8,835     3,916     4,085         822     617
75,796      9,596     3,915     4,085       1,586     759
```

Arena flat, memory cache at its configured budget, and all movement in `mmap`
outside the cache — which *oscillates* (811 → 1,318 → 811 → 1,272 → 822) rather
than ratcheting. That is in-flight tiles during prefetch bursts, returned each
time, which is the mmap threshold pin working as intended. One baseline shift of
~775 MB around 64,000 encodes is the only durable move, and it is a separate
question from this PR.

## Environment override

`mallopt` silently overrides what glibc read from the environment at startup, so
the override only works if we check for it. Measured on glibc 2.44, 64 threads:

```
env=64, no mallopt              : arena=23 MB
env=64, mallopt(4)  <- conflict : arena= 1 MB
```

Without the check, a user running an allocator experiment would have their
setting discarded with no diagnostic. Both pins now defer to an explicit
`MALLOC_*` variable **or** a `GLIBC_TUNABLES` entry naming the same parameter,
and log at INFO when they do.

This also fixes the pre-existing case for the mmap threshold, which checked only
`MALLOC_MMAP_THRESHOLD_` and would have silently overridden
`GLIBC_TUNABLES=glibc.malloc.mmap_threshold=...`.

## Verification

`make pre-commit` green. Two mutations checked, each caught only by its own test:

| mutation | test |
|---|---|
| drop the environment check | `an_explicit_env_setting_defers_our_pin`, `a_tunables_setting_defers_our_pin` |
| naive `tunables.is_some()` ignoring which key | `unrelated_tunables_do_not_defer_our_pin` |

The environment decision is a pure function, so the tests do not mutate
process-global state.

`configure_allocator()` now returns `AllocatorConfig` rather than `bool` — one
flag per pin, since a single bool cannot say which of two independent settings
applied. Both existing callers ignored the return value.

## Scope note

Both pins are `cfg`-gated to `linux-gnu`, as is all allocator telemetry. macOS
uses libmalloc and has neither. That gap is out of scope here and is the subject
of a separate v0.5.0 allocator evaluation.

## Status

**Flight in progress, ~2.5 h remaining.** This PR will be updated with end-of-
flight metrics — final arena drift, the crossover point, and total committed at
matched work. Not merge-ready until then.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,6 +250,20 @@ applied in code, not by environment variable, and both have flight evidence:
 - **Blocking pool cap** — tokio's 512 default let the pool reach 581 threads,
   and each excursion ratcheted glibc's arena high-water mark by ~420 MB. The
   arena reached 6,166 MB while holding 572 MB of live data.
+- **`mallopt(M_ARENA_MAX, 4)`** — glibc's default of `8 x ncores` (256 here)
+  lets every burst recruit fresh arenas, each keeping its own high-water mark
+  for the life of the process, so the total is the sum of every peak ever
+  seen and never converges. Capping the count removes the supply: the sum
+  saturates once each arena has seen its worst burst. On the reference leg the
+  arena settled at 3,916 MB within six minutes and held to the byte — peak
+  concurrent demand (3,093 MB) plus ~27% headroom — against 6,091 MB still
+  climbing at hour 10.75 uncapped. Override with `MALLOC_ARENA_MAX`.
+
+Both pins defer to an explicit `MALLOC_*` variable or `GLIBC_TUNABLES` entry:
+`mallopt` silently overrides what glibc read at startup, so without that check
+a user's own setting would vanish with no diagnostic. **Both are `cfg`-gated to
+`linux-gnu`** — macOS uses libmalloc and currently has neither these fixes nor
+the allocator telemetry.
 
 DDS payloads are `bytes::Bytes` end to end (#237): a tile is 10.66 MiB and a
 FUSE read wants ~4% of it, so cloning a payload costs more than delivering it.
--- a/xearthlayer-cli/src/main.rs
+++ b/xearthlayer-cli/src/main.rs
@@ -166,12 +166,24 @@ enum Commands {
 // ============================================================================
 
 fn main() -> ExitCode {
-    // First, before anything allocates in earnest. glibc's mmap threshold
-    // adapts upward past the 11.17 MiB DDS tile size after a single
-    // allocate/free cycle, after which tiles come from arenas and are never
-    // returned to the OS on free. Pinning it costs ~0.16 ms per tile and
-    // bounded arena retention at 4.6 MB against 191.5 MB in a 64-thread
-    // benchmark. Set MALLOC_MMAP_THRESHOLD_ to override. See issue #227.
+    // First, before anything allocates in earnest, and before any thread is
+    // created — the arena ceiling only bounds arenas not yet made.
+    //
+    // Two glibc parameters, both from issue #227:
+    //
+    // - mmap threshold: glibc adapts it upward past the 10.66 MiB DDS tile
+    //   size after a single allocate/free cycle, after which tiles come from
+    //   arenas and are never returned on free. Pinning costs ~0.16 ms per tile
+    //   and bounded arena retention at 4.6 MB against 191.5 MB in a 64-thread
+    //   benchmark.
+    // - arena ceiling: glibc's default of 8 x ncores lets every burst recruit
+    //   fresh arenas, each keeping its own high-water mark forever. An
+    //   11-hour flight reached a 6,091 MB arena holding 1,002 MB of live data,
+    //   still growing at hour 10.75. Capped, the same route settled at
+    //   3,916 MB within six minutes and held to the byte.
+    //
+    // Either can be overridden with MALLOC_MMAP_THRESHOLD_ / MALLOC_ARENA_MAX
+    // or GLIBC_TUNABLES; an explicit user setting always wins.
     xearthlayer::metrics::configure_allocator();
 
     let cli = Cli::parse();
--- a/xearthlayer/src/metrics/memory_probe.rs
+++ b/xearthlayer/src/metrics/memory_probe.rs
@@ -390,6 +390,58 @@ mod tests {
         assert_eq!(a.obtained_bytes(), 100_900);
     }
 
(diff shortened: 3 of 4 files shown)


samsoir:
## Flight complete — 11 h 18 min, 123,226 encodes, 31.5 M chunks

Same route as the baseline (FAOR→EGLL), slightly more work done (123,226 vs
121,641 encodes).

### End state

| | no cap (256 arenas) | `ARENA_MAX=4` |
|---|---|---|
| committed | 11,580 MB | **9,687 MB** |
| arena | 6,091 MB | **4,038 MB** |
| arena idle free list | 5,089 MB (84%) | 3,303 MB (82%) |
| swap | 1,063 MB | **533 MB** |
| encodes/sec | 3.06 | 3.03 |
| failed chunks | 113 | 0 |

**−1,893 MB committed (−16%), −2,053 MB of arena (−34%), half the swap.**
Throughput unchanged.

### Matched work

| encodes | no cap | `ARENA_MAX=4` | delta |
|---:|---:|---:|---:|
| 40,000 | 7,258 (arena 1,914) | 8,834 (arena 3,916) | +1,576 (+22%) |
| 60,000 | 8,182 (arena 2,819) | 8,830 (arena 3,915) | +648 (+8%) |
| 80,000 | 9,572 (arena 3,763) | 9,597 (arena 3,915) | +25 (0%) |
| 95,000 | 9,593 (arena 4,222) | 9,596 (arena 3,915) | +3 (0%) |
| 105,000 | 10,070 (arena 4,767) | 10,038 (arena 3,915) | −32 (0%) |
| 120,000 | 11,431 (arena 6,091) | 9,736 (arena 4,038) | **−1,695 (−15%)** |

**Crossover at ~98,000 encodes.** The cap front-loads the whole arena into the
first six minutes, so it is behind early; the uncapped run then accumulates past
it and keeps going.

### The ceiling held

```
arena min=3915  max=4040  drift=125 MB  over 672 samples
```

| | no cap | `ARENA_MAX=4` |
|---|---|---|
| steps > 100 MB | 21 | 7 |
| steps after initial sizing (>7,500 encodes) | 18 | **1** |
| total ratchet | 5,970 MB | 4,028 MB (3,903 of it in the first six minutes) |

Six steps sized the arena inside the first six minutes. **One** step followed in
the remaining 11 hours: +125 MB at `threads=205`, when a burst finally exceeded
every previous one. Uncapped took 18 post-sizing steps and was still stepping at
hour 10.75.

That is a **43× reduction in ratchet after sizing** (125 MB vs 5,393 MB), and it
is the difference between a bound and a slope.

## What this is, mechanically

An arena grows to cover the worst burst *it* ever saw and glibc never releases
that space, so `total arena = sum over arenas of (worst burst that arena saw)`.
With 256 arenas available, every burst can recruit fresh ones that have never
seen a peak — the sum is the running total of every burst across the flight,
which is why it never converged.

Capped at 4, only four high-water marks can exist. The total converges on
`4 × worst burst ever seen`, and once that burst has happened, identical events
cost nothing. The single late step is the ceiling being *refined* by a new worst
case, not a ratchet resuming.

Full mechanism and the operator observation that identified it: #227.

## Caveats

**This does not make memory flat.** Committed still rose from 8,834 at 40,000
encodes to 9,687 at the end. The arena is bounded; the remaining movement is in
`mmap`, which oscillates and returns (811 → 1,318 → 811 during bursts) with one
durable ~775 MB baseline shift. Different mechanism, tracked separately.

**The ceiling is our own concurrency, not an allocator setting.** 3,916 MB is
peak concurrent demand (3,093 MB measured) plus ~27% headroom. The peak is a
prefetch/boundary-load collision — filed as #246, and lowering it lowers this
floor directly.

**Linux only.** Both pins are `cfg`-gated to `linux-gnu`, as is all allocator
telemetry; macOS uses libmalloc and has neither. Out of scope here, tracked as
#245.

## Ready to merge

`make pre-commit` green, CI green, two mutations verified, no throughput
regression, zero failed chunks over 31.5 M.

PR #252, opened 2026-08-26
fix(diagnostics): bound the du call that could hang the report

### LINKED ISSUE #251: [Bug]: 'xearthlayer diagnostics' can hang indefinitely on an unbounded 'du' over the cache directory
### Bug Description

`xearthlayer diagnostics` can hang indefinitely. `report.rs:282-284` measures the cache directory with an unbounded `du -sh`:

```rust
if let Ok(output) = Command::new("du")
    .args(["-sh", &cache_dir.to_string_lossy()])
    .output()
```

`.output()` blocks until the child exits, and `du -sh` is a full recursive walk with a `stat()` per file. On a large cache in the region-based layout that is minutes at best. There is no timeout anywhere in the diagnostics module.

Three things make this worse than a merely slow command:

1. **It is the only unbounded collector.** The other five external commands in the same file — `uname` (`:117`), `lspci` (`:200`), `nvidia-smi` (`:214`), `df` (`:262`), `ip` (`:349`) — are all O(1). Only `du` scales with cache contents.

2. **The cheap technique is already in the same function.** Ten lines below, available space comes from `statvfs` via `crate::system::fs_info` (`system/filesystem.rs:47`) — one syscall. Only the *used* figure takes the expensive route.

3. **This command is mandatory for filing any bug.** `bug_report.yml` marks System Diagnostics as required and adds a required checkbox confirming the output is included. So the users with the largest caches — arguably the ones hitting the most interesting bugs — are the ones who cannot get past the template.

There is also a smaller correctness problem in the same area: when `du` fails, `cache_dir_size` stays `None` and the report prints `(not created)` (`report.rs:535`) for a directory that demonstrably exists, since the `du` call is inside an `if cache_dir.exists()` guard.

### Steps to Reproduce

1. Point `cache.directory` at a large cache (order of a terabyte, hundreds of thousands of files).
2. Run `xearthlayer diagnostics`.
3. The report stops after `## GPU Compute Adapters` and does not return. In my case it did not finish within three minutes; an earlier attempt ran for roughly an hour before being killed.

Contended I/O makes it worse — an `xearthlayer run` session on the same disk was active during my measurement.

### Expected Behavior

`xearthlayer diagnostics` terminates promptly and prints a complete report, degrading the cache size to an explicit "not measured" rather than blocking on it.

### Actual Behavior

The command hangs; no report is printed at all, so none of the fast, useful fields (version, kernel, GPU, available space) are reachable either.

### System Diagnostics

```text
XEarthLayer Diagnostics
=======================

XEarthLayer Version: 0.4.7-beta.1

## Operating System
Kernel: 7.1.10-2-liquorix-amd64
OS: Debian GNU/Linux 13 (trixie)
Desktop: KDE
Display Server: wayland

## Hardware
CPU: AMD Ryzen 7 9800X3D 8-Core Processor (16 cores)
Memory: 92.0 GB

## GPU
GPU: NVIDIA Corporation AD102 [GeForce RTX 4090] (rev a1)
GPU: Advanced Micro Devices, Inc. [AMD/ATI] Granite Ridge [Radeon Graphics] (rev cb)
VRAM (NVIDIA): 24564 MB (24.0 GB)

## GPU Compute Adapters
2026-08-26 11:58:42Z  INFO EGL says it can present to the window but not natively
  [0] NVIDIA GeForce RTX 4090 (Discrete, vulkan)
  [1] AMD Radeon Graphics (RADV RAPHAEL_MENDOCINO) (Integrated, vulkan)
  [2] llvmpipe (LLVM 19.1.7, 256 bits) (Software fallback, vulkan)
  [3] NVIDIA GeForce RTX 4090/PCIe/SSE2 (Other, gl)

## Disk Space
Cache directory: /mnt/xplane_data/xearthlayer/cache (1,5T used, 3241.8 GB available)
Packages location: /home/maechtel/.xearthlayer/packages (671.5 GB available)
Root filesystem: 171G used of 504G (36%)
Config directory: exists

## Network
Default interface: enp11s0

(config.ini dump intentionally omitted - contains provider API keys)
```

This block is itself truncated by the bug — it is what the command had emitted before it stopped returning.

### Relevant Logs

Not applicable.

### Severity

Minor functionality broken

---

### Suggested fix

Bound the child rather than dropping the field. PR incoming: a small `run_bounded(cmd, timeout)` helper that spawns, polls `try_wait()` and kills at the deadline, applied to the `du` call with a 5s budget, plus a marker so the line reads `(size unmeasured, 3241.8 GB available)` instead of `(not created)`.

Available space is untouched, since it never went through `du`.

On the cache above, `xearthlayer diagnostics` goes from "does not return" to 6 seconds end to end.

mmaechtel:
Fixes #251.

`xearthlayer diagnostics` measured the cache directory with an unbounded
`du -sh` (`report.rs:282-284`). Every other collector in that file is O(1);
this one walks the cache tree. On a 1.5 TB cache it did not return within
three minutes, so the report never printed — and since `bug_report.yml`
makes the diagnostics output a required field, that blocks filing a bug at
all.

### What this does

- Adds `run_bounded(cmd, timeout)`: spawns the child with piped stdout,
  polls `try_wait()`, and kills and reaps it at the deadline. A slow command
  now costs the caller the timeout instead of however long it would have
  taken. `None` covers every failure mode, which is what the collectors here
  already treat as "no value to report".
- Applies it to the `du` call with a 5s budget.
- Fixes a related mislabel: the `du` call sits inside an
  `if cache_dir.exists()` guard, so on failure the old `None` fallback
  printed `(not created)` for a directory that plainly exists. The line now
  reads `(size unmeasured, 3241.8 GB available)`.

No new dependency — `try_wait` plus a 25ms poll rather than pulling in
`wait-timeout`.

Available space is untouched: it already came from `statvfs`
(`system/filesystem.rs:47`), never from `du`.

### Scope

Only `du` is wrapped. The other five commands (`uname`, `lspci`,
`nvidia-smi`, `df`, `ip`) are O(1) and left as they are; `run_bounded` is
there if a future collector needs it.

### Tests

Six, written before the implementation:

- `run_bounded` — returns stdout on success; returns `None` at the deadline
  **and** proves it did not block (asserts elapsed < 5s against a `sleep 30`);
  returns `None` for a missing binary; returns `None` on non-zero exit.
- report formatting — a measured size still renders as `1.5T used`, and the
  unmeasured marker does not get ` used` appended to it.

### Measured

On the cache from #251, `xearthlayer diagnostics` goes from "does not return"
to 6 seconds end to end, with the rest of the report intact.

`make pre-commit` passes.

### DIFF
--- a/xearthlayer/src/diagnostics/report.rs
+++ b/xearthlayer/src/diagnostics/report.rs
@@ -3,7 +3,8 @@
 use std::fmt;
 use std::fs;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
+use std::time::{Duration, Instant};
 
 /// A comprehensive system diagnostics report.
 #[derive(Debug, Clone)]
@@ -279,17 +280,16 @@ impl DiskInfo {
         let cache_dir = &config.cache.directory;
         info.cache_dir_path = Some(cache_dir.display().to_string());
         if cache_dir.exists() {
-            if let Ok(output) = Command::new("du")
-                .args(["-sh", &cache_dir.to_string_lossy()])
-                .output()
-            {
-                if output.status.success() {
-                    let du = String::from_utf8_lossy(&output.stdout);
-                    if let Some(size) = du.split_whitespace().next() {
-                        info.cache_dir_size = Some(size.to_string());
-                    }
-                }
-            }
+            let mut du = Command::new("du");
+            du.args(["-sh", &cache_dir.to_string_lossy()]);
+            // The directory exists, so a missing size means `du` did not finish
+            // in time (or failed) - not that there is nothing there. Say so
+            // rather than letting it read as "(not created)".
+            info.cache_dir_size = Some(
+                run_bounded(du, COMMAND_TIMEOUT)
+                    .and_then(|out| out.split_whitespace().next().map(str::to_string))
+                    .unwrap_or_else(|| UNMEASURED_SIZE.to_string()),
+            );
         }
         if let Ok(fs) = crate::system::fs_info(cache_dir_for_statvfs(cache_dir)) {
             info.cache_dir_available =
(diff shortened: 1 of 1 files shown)


xearthlayer/src/diagnostics/report.rs
```diff
@@ -406,6 +406,61 @@ fn parse_section_header(line: &str) -> Option<&str> {
...
+                    return None;
+                }
+                std::thread::sleep(COMMAND_POLL_INTERVAL);
+            }
+            Err(_) => return None,
+        }
+    }
+}
+
 fn redact_sensitive_ini(content: &str) -> String {
```
samsoir, line 476: Doc comment on this method is missing. credential-redaction block referencing issue #162 now documents UNMEASURED_SIZE, a cache-size string.

xearthlayer/src/diagnostics/report.rs
```diff
@@ -406,6 +406,61 @@ fn parse_section_header(line: &str) -> Option<&str> {
...
+/// Runs `cmd` and returns its stdout, or `None` if it fails or outlives
+/// `timeout`.
+///
+/// On timeout the child is killed and reaped, so a slow command costs the
+/// caller `timeout` rather than however long the command would have taken.
+/// `None` means "no value to report" for every failure mode — the collectors
+/// here all treat a missing field as a normal outcome.
+fn run_bounded(mut cmd: Command, timeout: Duration) -> Option<String> {
+    let mut child = cmd
+        .stdout(Stdio::piped())
```
samsoir, line 437: pipes stdout and polls try_wait without draining. Fine for du -sh, but it's written and documented as a general helper. Worth a doc line since the signature invites reuse, or consider refactoring to allow for general usage.

samsoir (dismissed):
Thanks for the PR and fixing these issues. Few minor comments before ready to merge.

xearthlayer/src/diagnostics/report.rs
samsoir: Fixed in f353b9c — moved the nine-line block back to sit immediately above `fn redact_sensitive_ini`, leaving `UNMEASURED_SIZE` with only its own doc. Pushed to this branch directly so it can go out in beta.2 tonight rather than waiting on a round-trip; shout if you would rather have made the change yourself.

xearthlayer/src/diagnostics/report.rs
samsoir: Documented rather than refactored, in f353b9c.

Reasoning, since you offered both: `run_bounded` is module-private with a single caller that reports one line, so draining concurrently would mean carrying a reader thread for a reuse case that does not exist yet. The current behaviour is also safe rather than merely undocumented — if the pipe fills, the child blocks, the deadline still fires and the child is still killed. The cost is that the call always burns the full timeout and always returns `None`, which *reads* like a slow command rather than a full pipe. That is the part worth writing down, so the doc block now names the 64 KiB bound, the exact symptom, and says draining is the right change once a chatty caller needs it.

Happy to take the refactor instead if you would rather have it general now — it is a small change, just not one tonight.

samsoir:
:shipit:

Issue #233, opened 2026-08-19
FUSE read() reads the entire file to serve a ≤1 MiB range


samsoir:
## Problem

Every FUSE `read()` of a real passthrough file reads the **entire file** from disk into a
fresh `Vec<u8>`, returns the requested slice, and drops the rest. The kernel never asks for
a whole file: it caps every FUSE read request at `max_pages × PAGE_SIZE`, which is 1 MiB at
most. Measured against a purpose-built probe FS on our own `fuse3` fork, a single whole-file
`read(2)` of a 32 MiB file arrives as **32 separate 1 MiB read calls**, and an ordinary
sequential reader produces 512 KiB calls. So the handler rebuilds the whole object once per
call: reading an 11.17 MB file costs 12–23 full reads and allocations of that size, and the
largest installed ortho DSF (237.5 MB) costs up to 238.

The cost is mostly memory bandwidth and allocation churn rather than disk I/O — after the
first pass the underlying file is in the host page cache — but the allocations are large,
repeated, and land in the size band implicated by #227. They are also gated by the FUSE disk
I/O limiter (`min(num_cpus × 16, 256)` permits) rather than by `max_concurrent_jobs`, so
concurrent scene loading can hold hundreds of full-file buffers at once. A secondary effect
is first-byte latency: nothing is returned until the whole file has been read, so X-Plane
waits on a full 237 MB read to receive its first 1 MiB.

Affected files served this way include ortho DSFs (18,150 files, mean 16.5 MB, max 237.5 MB),
patch DDS textures (9,255 files, mean 13.12 MB) and `.ter` files.

## Sites

- `xearthlayer/src/fuse/fuse3/ortho_union_fs.rs:784`
- `xearthlayer/src/fuse/fuse3/union_fs.rs:424`
- `xearthlayer/src/fuse/fuse3/filesystem.rs:374`

## Notes

Independent of kernel page caching — this is about what the handler does once a call arrives,
not about whether the kernel caches between calls. #65's `FOPEN_DIRECT_IO` decision for
virtual DDS files is unaffected, and per-read telemetry (`FuseLoadMonitor`, `SceneTracker`,
`DdsAccessEvent`) continues to see every read.

## Related

#227 (unbounded memory growth) · #234 (the virtual-DDS twin of this defect; to be fixed in the same pass) · #65 · #204

samsoir:
## Scope addition: FUSE read metrics

[...]

PR #264, opened 2026-09-07
Merge main (v0.4.8) into develop/0.5.0

samsoir:
Catch-up merge. `develop/0.5.0` was 161 behind main and 41 ahead, last touched 2026-08-04.

Brings across the #227 memory work, #226 region states, #233/#234 FUSE read amplification, the cache integrity model, and the 0.4.7 and 0.4.8 releases. The 41 commits already here are dlicudi's macOS port (#202).

## Approach

**Main's composition takes precedence.** The macOS port's *behaviour* is preserved, but ported onto main's structure rather than restoring the structure it was written against.

16 conflicts. Eleven were version strings and documentation. Five needed judgement:

**`fuse3/filesystem.rs` deleted.** #233 made it unreachable, so the macOS fix that lived there (`d4275a3`) had been applied to a file that no longer exists. Taking the deletion alone would have silently dropped the macOS mmap-safety rule.

That rule is preserved as a `cfg` split on the file-local constant in `ortho_union_fs.rs`: `FOPEN_DIRECT_IO` on Linux, `0` on macOS, because X-Plane mmaps texture files and macFUSE faults with `EXC_BAD_ACCESS` on a `direct_io` mmap. Develop's `shared.rs` additions are reverted; they existed to stop the flag drifting between two filesystems and there is now one.

**`config/storage.rs` deleted.** Main removed it alongside `cache.disk_io_profile` (`2145e1e`) after confirming nothing consumed the sizing, and moved detection to `system/storage_detect.rs`. Develop's macOS `df` + `diskutil` detection is ported there, retargeted from `DiskIoProfile` to `StorageType`, with its seven tests.

**`hardware.rs`** takes develop's memory handling, which flags a failed detection rather than presenting the 8 GB fallback as real (`07ec9dc`), and main's storage lookup.

**`Makefile`** takes develop's `bump-version`. It is a strict superset: pre-release identifiers via `BASE_VERSION`, `version.json` through `jq`, BSD-portable sed, and the Fedora tag read back rather than hardcoded.

**`Cargo.lock`** is rebased on develop's rather than main's, because develop repointed the `fuse3` fork to `master` while main's lock pinned `configurable-background`. Main's `memory-stats` and `csscolorparser` were then resolved in.

## Two silent auto-merges corrected

Neither produced a conflict marker, and both would have passed CI.

`shared.rs` ended up with `FOPEN_DIRECT_IO` defined twice, once file-local from main and once `pub` from develop, under a doc comment claiming a single source of truth. Main deleted it from `shared.rs` and re-declared it elsewhere; develop enhanced it in place; the edits never touched the same lines.

The storage case is the same shape one level up. Git flagged it only because main deleted the file outright rather than gutting it.

## Also

Two pre-existing staleness bugs fixed in files already being resolved: `shared.rs` still documented two filesystems #233 deleted, and `hardware.rs`'s `storage_type` doc still referenced the removed SSD profile.

## Verification

`make pre-commit` green on Linux: 2538 passed, 0 failed, across 11 suites. That is 10 more tests than main, so develop's additions came through.

**Linux CI cannot validate this merge.** The cfg'd macOS paths are never type-checked here, and the ported `storage_detect` and open-flags changes are exactly what a Mac exercises. `Verify (macOS)` on `macos-15` is the first real check, and a live macFUSE mount test is still needed beyond that, since hosted runners cannot load a kernel extension.


samsoir:
@dlicudi this will require some work to get stable on macOS again. I will be testing locally, but any help verifying it from your end will be valuable if you have available cycles.

samsoir:
Verified as working and green on local mac studio.

PR #273, opened 2026-09-13
fix(diagnostics): report the config file that is actually in use

mmaechtel:
## Problem

`xearthlayer diagnostics` names and prints the wrong configuration file on any
installation that has been migrated to the 0.5.0 layout.

`ConfigInfo::collect()` constructed its path directly:

```rust
let home = dirs::home_dir().unwrap_or_default();
let config_path = home.join(".xearthlayer/config.ini");
```

The migration deliberately leaves `~/.xearthlayer/config.ini` in place — it
carries a "this file is no longer read" banner so that an older binary still
finds a working installation. Because the collector never asks the path
resolver, it finds that leftover first and reports it as the configuration in
force. Observed on a migrated install (`layout_version = 1`, `migrate status`
reporting nothing pending):

```
Config file: /home/<user>/.xearthlayer/config.ini

; This file is no longer read by XEarthLayer 0.5.0-dev.0 or later.
; The configuration in use is:
;   /home/<user>/.config/xearthlayer/config.ini
```

`docs/getting-started.md` and the bug report template both point users at
`xearthlayer diagnostics`, so every report from a migrated installation would
have carried settings XEarthLayer never loaded.

## Fix

- `ConfigInfo::from_path(&Path)` does the read; `collect()` supplies
  `paths::config_file()`. This restores the rule from `CLAUDE.md` that the
  config path is only ever obtained from a role accessor.
- Three doc comments still named `~/.xearthlayer/packages` as the install
  default. That fallback has gone through `paths::packages_dir()` since #269 —
  only the comments were stale.

## Tests

Two new unit tests, written first:

- `config_info_reads_the_file_it_is_given` — builds a config under
  `TestDirectories`, resolves it with `paths::config_file_in()`, asserts the
  report names that file and contains its contents.
- `config_info_is_empty_when_the_file_does_not_exist` — absent file yields an
  empty report rather than a path with no contents.

Both use the `BaseDirectories` test double, so neither reads a real home
directory. `make pre-commit` passes.

https://claude.ai/code/session_0122xyYevvFu6bnuJbdM6Ahp


samsoir:
lgtm :shipit:

